PR #51, opened 2023-05-22
Feat/config builders

l0r1s:
- [x] Add configuration
- - [x] Relaychain
- - [x] Parachain 
- - [x] HRMP channels
- - [x] Network

- [x] Add builders
- - [x] Relaychain
- - [x] Parachain 
- - [x] HRMP channels
- - [x] Network

- [x] Add tests for builders
- - [x] Relaychain
- - [x] Parachain 
- - [x] HRMP channels
- - [x] Network

- [ ] Add validation
- - [ ] Relaychain
- - [ ] Parachain 
- - [ ] HRMP channels
- - [ ] Network

- [ ] Add test for validation
- - [ ] Relaychain
- - [ ] Parachain 
- - [ ] HRMP channels
- - [ ] Network


crates/configuration/src/network.rs
```diff
@@ -0,0 +1,92 @@
...
+    relaychain: RelaychainConfig,
+
+    /// Parachains configurations.
+    parachains: Vec<ParachainConfig>,
+
+    /// HRMP channels configurations.
+    hrmp_channels: Vec<HrmpChannelConfig>,
+
+    // [TODO]: what does it represents ?
+    config_base_path: String,
```
pepoviola, line 47: This represent the `base path of the config file`, we don't need it here.

pepoviola, line 47:
```suggestion
    // [TODO]: what does it represents ?
    config_base_path: String,
```

crates/configuration/src/network.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/// Global settings applied to an entire network.
+pub struct GlobalSettings {
+    /// Whether we should spawn a dedicated bootnode for each chain.
+    spawn_bootnode: bool,
+
+    /// External bootnode address.
+    /// [TODO]: is it a default overriden by node config, maybe an option ?
+    bootnode_address: MultiAddress,
```
pepoviola, line 13: Yes, the expected behavior is that this bootnode is added to the `relaychain` spec. But we can drop this field and revisit later. We don't have any use-case that use this feature.

```suggestion
    /// External bootnode address.
    /// [TODO]: is it a default overriden by node config, maybe an option ?
    bootnode_address: MultiAddress,
```

crates/configuration/src/network.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Global spawn timeout in seconds.
+    spawn_timeout: Timeout,
```
pepoviola, line 16:
```suggestion
    network_spawn_timeout: Timeout,
```
I think is a better naming.

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Chain to use (use None if you are running adder-collator or undying-collator).
+    chain: Option<String>,
+
+    /// Wether to add the parachain to the genesis (chain specification) file.
+    has_to_be_added_to_genesis: bool,
+
+    /// Wether to register this parachain (via genesis or extrinsic).
+    /// [TODO]: is the "via genesis" part of the comment needed given the above option add to genesis ?
+    has_to_be_registered: bool,
```
pepoviola, line 20: I think we can merge this options and use an enum. The possible options are:
- Register the para in the genesis
- Register the para using and extrinsic
- Not register the para.

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Parachain balance.
+    /// [TODO]: rename to initial_balance ? shouldnt be u128 ?
```
pepoviola, line 23: Yes, to both.

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+    balance: u64,
+
+    /// Path to WASM runtime.
+    genesis_wasm_path: Option<Path>,
```
pepoviola, line 27: We want to support specify a `path` or an `url`.

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Command to generate the WASM runtime.
+    genesis_wasm_generator: Option<Command>,
+
+    /// Path to the gensis `state` file.
+    genesis_state_path: Option<Path>,
```
pepoviola, line 33: We want to support specify a `path` or an `url`.

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Command to generate the genesis `state`.
+    genesis_state_generator: Option<Command>,
+
+    /// Use a pre-generated chain specification.
+    chain_spec_path: Option<Path>,
```
pepoviola, line 39: We want to support specify a `path` or an `url`.

crates/configuration/src/shared/types.rs
```diff
@@ -0,0 +1,93 @@
+#[derive(Debug, Clone)]
+pub struct MultiAddress(String);
+
+#[derive(Debug, Clone)]
+pub struct IpAddress(String);
+
+#[derive(Debug, Clone)]
+pub struct Timeout(String);
```
pepoviola, line 8: We should use something like `Duration`.

crates/configuration/src/shared/types.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn with_limit_cpu(self, quantity: ResourceQuantity) -> Self {
+        Self {
+            limit_cpu: Some(quantity),
+            ..self
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum DbSnapshot {
```
pepoviola, line 67:
```suggestion
pub enum FileOrUrl {
```

crates/configuration/src/network.rs
```diff
@@ -0,0 +1,92 @@
+use crate::{
+    shared::types::{IpAddress, MultiAddress, Timeout},
+    HrmpChannelConfig, ParachainConfig, RelaychainConfig,
+};
+
+/// Global settings applied to an entire network.
+pub struct GlobalSettings {
```
wirednkod, line 7: This does not have to be public (at least as of this code):
```suggestion
struct GlobalSettings {
```

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl Default for ParachainConfig {
+    fn default() -> Self {
+        // [TODO]: define the default value for a parachain
+        todo!()
+    }
+}
+
+impl ParachainConfig {
+    pub fn with_id(self, id: u16) -> Self {
```
wirednkod, line 63: I think its shorter and preferable to rename all `with_` to `w_`

crates/configuration/src/relaychain.rs
```diff
@@ -0,0 +1,118 @@
...
+            ..self
+        }
+    }
+
+    pub fn with_default_db_snapshot(self, db_snapshot: DbSnapshot) -> Self {
+        Self {
+            default_db_snapshot: Some(db_snapshot),
+            ..self
+        }
+    }
```
wirednkod, line 78: I do not think that there are default values; Meaning if one do not supply then there is no "default replacement"

crates/configuration/src/network.rs
l0r1s: We can't else are exposing the private API in the following method:

```rust
    pub fn with_global_settings(self, f: fn(GlobalSettings) -> GlobalSettings) -> Self {
        Self {
            global_settings: f(self.global_settings),
            ..self
        }
    }
```

crates/configuration/src/relaychain.rs
l0r1s: There is an implementation of the ```Default``` trait to be determined so we can know what we expect as good defaults:

```rust
impl Default for RelaychainConfig {
    fn default() -> Self {
        // [TODO]: define the default value for a relaychain
        todo!()
    }
}
```

crates/configuration/src/network.rs
wirednkod: Im not sure I understand. This method exists in the same file.
What I am trying to say, is that everything should be init as private until they are needed to be declared as public. Its better than declare everything by default as public

crates/configuration/src/parachain.rs
wirednkod: as far as I remember `w_` at the beginning always means with... IMO since `with_` exists in all the `fns` it makes it verbose and unnecessery.

crates/configuration/src/relaychain.rs
wirednkod: I mean - there cannot be a default image, when not provided. Or a default command. User always needs to provide those. the only default value I remember is the `cumulus_based` which defaults to `true`

crates/configuration/src/network.rs
l0r1s: The method exists in the same file but is public, this is the API we will expose to the user. Every struct field is private but every method on the builder is public else the user won't be able to fill some fields due to the field being private.

crates/configuration/src/shared/types.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct MultiAddress(String);
+
+#[derive(Debug, Clone)]
+pub struct IpAddress(String);
+
+#[derive(Debug, Clone)]
+pub struct Timeout(String);
+
+#[derive(Debug, Clone)]
+pub struct Command(String);
```
pepoviola, line 11: We have a couple of options to use these `new-type` in the `configuration`, we can change the field visibility (e.g `pub(crate)` ) or we can add a constructor (e.g `pub fn new(cmd: impl Into<String>`))

I think the later is more appropriated but I'm open to discuss other ideas.

crates/configuration/src/global_settings.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    pub fn local_ip(&self) -> Option<&IpAddress> {
+        self.local_ip.as_ref()
+    }
+}
+
+#[derive(Debug)]
+pub struct GlobalSettingsBuilder {
+    config: GlobalSettings,
```
pepoviola, line 43: Do we need the inner `config` here? what's the benefit to use it?

crates/configuration/src/network.rs
```diff
@@ -0,0 +1,260 @@
...
+};
+
+/// A network configuration, composed of a relaychain, parachains and HRMP channels.
+#[derive(Debug, Clone, PartialEq)]
+pub struct NetworkConfig {
+    // The global settings applied to the network.
+    global_settings: GlobalSettings,
+
+    /// Relaychain configuration.
+    relaychain: Option<RelaychainConfig>,
```
pepoviola, line 18: I understand that the builder will `drive` the users to set the `relaychain` here, but I think we can drop the inner config and make two `similar` structs for `NetworkConfig` / `NetworkConfigBuider` making `optional` in the latest. wdyt?

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,345 @@
...
+pub enum RegistrationStrategy {
+    InGenesis,
+    UsingExtrinsic,
+}
+
+/// A parachain configuration, composed of collators and fine-grained configuration options.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ParachainConfig {
+    // Parachain ID to use.
+    id: u16,
```
pepoviola, line 19:
```suggestion
    id: u32,
```

crates/configuration/src/relaychain.rs
```diff
@@ -0,0 +1,302 @@
...
+    chain_spec_path: Option<AssetLocation>,
+
+    /// Default arguments to use in nodes. Can be overriden on each node.
+    default_args: Vec<Arg>,
+
+    /// Set the count of nominators to generator (used with PoS networks).
+    random_nominators_count: Option<u32>,
+
+    /// Set the max nominators value (used with PoS networks).
+    max_nominations: Option<u16>,
```
pepoviola, line 38:
```suggestion
    max_nominations: Option<u8>,
```

crates/configuration/src/relaychain.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    pub fn default_args(&self) -> Vec<&Arg> {
+        self.default_args.iter().collect::<Vec<&Arg>>()
+    }
+
+    pub fn random_minators_count(&self) -> Option<u32> {
+        self.random_nominators_count
+    }
+
+    pub fn max_nominations(&self) -> Option<u16> {
```
pepoviola, line 77:
```suggestion
    pub fn max_nominations(&self) -> Option<u8> {
```

pepoviola (changes requested):
Looks good! Couple of small changes and I think we have room to simplify some things. Awesome progress, I think we should merge and keep building on-top.
Thanks!

crates/configuration/src/parachain.rs
```diff
@@ -0,0 +1,345 @@
...
+pub struct ParachainConfigBuilder<State> {
+    config: ParachainConfig,
+    _state: PhantomData<State>,
+}
+
+impl Default for ParachainConfigBuilder<Initial> {
+    fn default() -> Self {
+        Self {
+            config: ParachainConfig {
+                id: 0,
```
pepoviola, line 121:
```suggestion
                id: 100,
```

crates/configuration/src/global_settings.rs
l0r1s: Here, it forces us to initialize the config through a builder, the only way to construct the GlobalSettings is through the builder ```.build()``` method so we can control how we construct the settings and add validation check in the ```.build()``` to ensure that when we have a GlobalSettings in the code, it's fully validated (for bootnodes addresses and local_ip)

crates/configuration/src/network.rs
pepoviola: Resolved, since we already talked.

pepoviola (approved):

PR #58, opened 2023-05-24
feature/providers - client (and native) builder

### LINKED ISSUE #37: Native (`./crates/providers/native` - features = "native")
Component of [Provider Crate](#25)

wirednkod:
Fixes #37


crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+            tmp_dir:               tmp_dir.to_owned(),
+            command:               "bash".to_owned(),
+            pod_monitor_available: false,
+            local_magic_file_path: format!("{}/finished.txt", tmp_dir),
+            provider_name:         "native".to_owned(),
+        }
+    }
+}
+
+impl Client for Native {
```
l0r1s, line 50: To create the implementation of an async trait, you only need to add the macro on top of the implementation, it avoids writing the lifetime to each method and the future return type of the method

```suggestion

#[async_trait]
impl Client for Native {
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
+use std::error::Error;
+
```
l0r1s, line 2:
```suggestion
use std::error::Error;
use async_trait::async_trait;
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+impl Client for Native {
+    fn create_namespace<'async_trait>() -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 59:
```suggestion
    async fn create_namespace() -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        settings: Settings,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 71:
```suggestion
    async fn static_setup(settings: Settings) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    fn destroy_namespace<'async_trait>() -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 81:
```suggestion
    async fn destroy_namespace() -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        with_timestamp: Option<bool>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<String, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 95:
```suggestion
    async fn get_node_logs(
        pod_name: String,
        since: Option<u32>,
        with_timestamp: Option<bool>,
    ) -> Result<String, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        pod_name: String,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 108:
```suggestion
    async fn dump_logs( path: String, pod_name: String) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        minutes: u32,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 120:
```suggestion
    async fn upsert_cron_job(minutes: u32) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        namespace: Option<String>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<u16, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 134:
```suggestion
    async fn start_port_forwarding(
        port: u16, identifier: String, namespace: Option<String>,
    ) -> Result<u16, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        opts: RunCommandOptions,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<RunCommandResponse, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 147:
```suggestion
    async fn run_command(
        args: Vec<String>,
        opts: RunCommandOptions,
    ) -> Result<RunCommandResponse, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        args: Vec<String>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<RunCommandResponse, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 161:
```suggestion
    async fn run_script(
        identifier: String,
        script_path: String,
        args: Vec<String>,
    ) -> Result<RunCommandResponse, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        db_snapshot: Option<String>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 177:
```suggestion
   async fn spawn_from_def(
        pod_def: PodDef,
        files_to_copy: Option<Vec<FileMap>>,
        keystore: Option<String>,
        chain_spec_id: Option<String>,
        db_snapshot: Option<String>,
    ) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        container: Option<String>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 192:
```suggestion
    async fn copy_file_from_pod(
        identifier: String,
        pod_file_path: String,
        local_file_path: String,
        container: Option<String>,
    ) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
```
l0r1s, line 205:
```suggestion
    async fn put_local_magic_file(
        name: String,
        container: Option<String>,
    ) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        wait_ready: bool,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 219:
```suggestion
    async fn create_resource(
        resourse_def: NameSpaceDef,
        scoped: bool,
        wait_ready: bool,
    ) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        chain: String,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 232:
```suggestion
    async fn create_pod_monitor(
        file_name: String,
        chain: String,
    ) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    fn is_pod_monitor_available<'async_trait>() -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), bool>> + core::marker::Send + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 250:
```suggestion
    async fn is_pod_monitor_available() -> Result<(), bool>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+    fn restart_node<'async_trait>(
+        name: String,
+        timeout: u32,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), bool>> + core::marker::Send + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 269:
```suggestion
    async fn restart_node(name: String, timeout: u32) -> Result<(), bool>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        port: Option<u16>,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<Vec<(String, u32)>, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 282:
```suggestion
    async fn get_node_info(
        identifier: String,
        port: Option<u16>,
    ) -> Result<Vec<(String, u32)>, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        identifier: String,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<String, Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 294:
```suggestion
    async fn get_node_ip(
        identifier: String,
    ) -> Result<String, Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        ws_uri: String,
+    ) -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), Box<dyn Error>>>
+                + core::marker::Send
+                + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 306:
```suggestion
    async fn spawn_intro_spector(ws_uri: String) -> Result<(), Box<dyn Error>> {
        todo!()
    }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    fn validate_access<'async_trait>() -> core::pin::Pin<
+        Box<
+            dyn core::future::Future<Output = Result<(), bool>> + core::marker::Send + 'async_trait,
+        >,
+    > {
+        todo!()
+    }
```
l0r1s, line 314:
```suggestion
    fn validate_access() -> Result<(), bool> {
        todo!()
    }
```

crates/provider/src/shared/client.rs
```diff
@@ -0,0 +1,71 @@
...
+    async fn restart_node(name: String, timeout: u32) -> Result<(), bool>;
+    async fn get_node_info(
+        identifier: String,
+        port: Option<u16>,
+    ) -> Result<Vec<(String, u32)>, Box<dyn Error>>;
+    async fn get_node_ip(identifier: String) -> Result<String, Box<dyn Error>>;
+    async fn spawn_intro_spector(ws_uri: String) -> Result<(), Box<dyn Error>>;
+    async fn validate_access() -> Result<(), bool>;
+    fn get_logs_command(name: String) -> String;
+}
```
l0r1s, line 71: I think we should name the main trait Provider because the create is a provider and we have different types of provider PodmanProvider, KubernetesProvider, NativeProvider...

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct NsMetadata {
+    name:   String,
+    labels: NsLabels,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct NameSpaceDef {
```
l0r1s, line 65:
```suggestion
pub struct NamespaceDef {
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct NsLabels {
+    job_id:       String,
+    project_name: String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct NsMetadata {
```
l0r1s, line 59:
```suggestion
struct NamespaceMetadata {
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct RunCommandOptions {
+    resource_def: Option<String>,
+    scoped:       Option<bool>,
+    allow_fail:   Option<bool>,
+    main_cmd:     String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct NsLabels {
```
l0r1s, line 53:
```suggestion
struct NamespaceLabels {
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+    job_id:       String,
+    project_name: String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct NsMetadata {
+    name:   String,
+    labels: NsLabels,
```
l0r1s, line 61:
```suggestion
    labels: NamespaceLabels,
```

l0r1s:
I'm not sure we should have a 1 to 1 mapping with the method in typescript, for example, there are some obsure things which seems hacky, for example, the "magic file". wdyt @pepoviola  ?

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Clone, PartialEq)]
+struct Port {
+    container_port: u16,
+    name:           PortName,
+    flag:           String,
+    host_port:      u16,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct GlobalVolume {
```
l0r1s, line 118:
```suggestion
struct Volume {
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+    name:       String,
+    fs_type:    String,
+    mount_path: String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Settings {
+    global_volumes:                     Option<Vec<GlobalVolume>>,
```
l0r1s, line 126:
```suggestion
    global_volumes:                     Option<Vec<Volume>>,
```

l0r1s:
For standard error definition, we could use  https://docs.rs/anyhow/latest/anyhow/ with https://docs.rs/thiserror/latest/thiserror/ which are heavily used in Rust and simplify grouping of all errors

crates/provider/src/docker.rs
```diff
@@ -0,0 +1,11 @@
+#[allow(dead_code)]
```
pepoviola, line 1: We can just remove this `template` files for now.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
+use std::error::Error;
+
+use crate::shared::{
+    client::Client,
+    types::{FileMap, NameSpaceDef, PodDef, RunCommandOptions, RunCommandResponse, Settings},
+};
+
+struct Native {
+    // Namespace of the client
```
pepoviola, line 9:
```suggestion
    /// Namespace of the client
```
We should use `///` for [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments)

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+
+impl Default for Native {
+    fn default() -> Self {
+        // [TODO]: define the default value for Native
+        todo!()
+    }
+}
+
+impl Native {
+    pub fn new(namespace: &str, config_path: &str, tmp_dir: &str) -> Native {
```
pepoviola, line 35:
```suggestion
    pub fn new(namespace: impl Into<String>, config_path: impl Into<String>, tmp_dir: impl Into<String>) -> Native {
```
Using `impl Into<String>` give us more flexibility.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,323 @@
...
+        Native {
+            namespace:             namespace.to_owned(),
+            config_path:           config_path.to_owned(),
+            debug:                 true,
+            timeout:               60, // seconds
+            tmp_dir:               tmp_dir.to_owned(),
+            command:               "bash".to_owned(),
+            pod_monitor_available: false,
+            local_magic_file_path: format!("{}/finished.txt", tmp_dir),
+            provider_name:         "native".to_owned(),
```
pepoviola, line 45: I don't think we need this field.
```suggestion
```

crates/provider/src/shared/client.rs
```diff
@@ -0,0 +1,71 @@
+use std::error::Error;
+
+use async_trait::async_trait;
+
+use super::types::{FileMap, NameSpaceDef, PodDef, Settings};
+use crate::shared::types::{RunCommandOptions, RunCommandResponse};
+
+#[async_trait]
+#[allow(non_upper_case_globals)]
+pub trait Client {
```
pepoviola, line 10:
```suggestion
pub trait Provider {
```

I think we need to call it `Provider`, since at the moment we have a wording mix between Provider and Client. I think a good way to reason about the naming is that each `provider` needs to `impl` the `Provider` trait and then interact with a `client` to deploy/manage the internal resources (e.g pods/process).
wdyt?

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+enum ImagePullPolicy {
+    IfNotPresent,
+    Never,
+    Always,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct FileMap {
+    local_file_path:  String,
+    remote_file_path: String,
```
pepoviola, line 32:
```suggestion
    local_file_path:  PathBuf,
    remote_file_path: PathBuf,
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[derive(Debug, Clone, PartialEq)]
+struct GlobalVolume {
+    name:       String,
+    fs_type:    String,
+    mount_path: String,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Settings {
```
pepoviola, line 125: We should revisit all this options.

crates/provider/src/shared/client.rs
pepoviola: I think we also have the opportunity here to refactor and have a more clear boundaries between the `provider` and the `networkNode` (since the responsibility of those are a little mix now).

wirednkod: Agreed @l0r1s ;
@pepoviola can you elaborate a bit more on this?

pepoviola: Yes, this was more in the spirit of trigger the conversation on how we want to design the interaction between those two `mods`, no need to be addressed/resolved as part of this pr.
In the `ts` version the `client` is used as singleton from the `networkNode` and I think we can improve that design in the new version to have a more clear/loose couple approach.

crates/provider/src/shared/client.rs
wirednkod: I agree - already changed it from @l0r1s 's comment. Thank you

crates/provider/src/shared/types.rs
wirednkod: What do you mean?

l0r1s: Going through each option and see if it makes sense to keep it in the settings

wirednkod: I think this is part of the "next stage" (meaning when we start to actually implement the client)

crates/provider/src/native.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    fn run_command(
+        &self,
+        mut args: Vec<String>,
+        opts: RunCommandOptions,
+    ) -> Result<RunCommandResponse, Box<dyn Error>> {
+        if let Some(pos) = args.iter().position(|x| *x == "bash") {
+            args.remove(pos);
+        }
```
pepoviola, line 110: we only want to remove `bash` if is the first item in the vector.
```suggestion
        if let Some(arg) = args.get(0) {
            if arg == "bash" { args.remove(0) }
        }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,271 @@
...
+
+        let output: Output = if cfg!(target_os = "windows") {
+            Command::new("cmd")
+                .args(args)
+                .output()
+                .expect("failed to execute process")
+        } else {
+            if let Some(pos) = args.iter().position(|x: &String| *x == "-c") {
```
pepoviola, line 118: Same here, only remove if is the first item in the vector.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,271 @@
...
+
+            Command::new("sh")
+                .arg("-c")
+                .arg(args.join(" "))
+                .output()
+                .expect("failed to execute process")
+        };
+
+        if opts.allow_fail.is_some() && opts.allow_fail.unwrap() {
+            panic!("{}", String::from_utf8(output.stderr).unwrap());
```
pepoviola, line 130: We should not `panic` as part of the sdk, also the idea here is that if `allow_fail` is set to true we allow to fail without return an error.

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,172 @@
...
+impl RunCommandOptions {
+    pub fn default() -> Self {
+        Self {
+            resource_def: None,
+            scoped:       None,
+            allow_fail:   None,
+            main_cmd:     String::new(),
+        }
+    }
+}
```
pepoviola, line 75:
```suggestion
```
we can just derive Default here.

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,172 @@
...
+    local_file_path:  PathBuf,
+    remote_file_path: PathBuf,
+    unique:           bool,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct RunCommandResponse {
+    pub exit_code: ExitStatus,
+    pub std_out:   Vec<u8>,
+    pub std_err:   Option<Vec<u8>>,
```
pepoviola, line 45: Why don't use `String`/Option<String>` here and convert from the output of the process?

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,172 @@
...
+    env:       ProcessEnvironment,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct PodDef {
+    metadata: PodMetadata,
+    spec:     PodSpec,
+}
+
+type ProcessEnvironment = BTreeMap<String, String>;
```
pepoviola, line 127: I think we should use a `HashMap` here, we don't need to ensure the order here.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,271 @@
...
+    namespace:             String,
+    /// path where configuration relies
+    config_path:           String,
+    // variable that shows if debug is activated
+    debug:                 bool,
+    /// the timeout for the client to exit
+    timeout:               u32,
+    /// command sent to client
+    command:               String,
+    /// temporary directory
```
l0r1s, line 41: Just my 2 cents but we should stay uniform in the doc comment, starting with uppercase + I don't think it's need to be doc domment given fields are private

crates/provider/src/shared/types.rs
wirednkod: The type of std::process::Output is : `pub stderr: Vec<u8, Global>`;

Isn't it easier to use that type instead of getting into conversions?

crates/provider/src/native.rs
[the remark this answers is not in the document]
wirednkod: Already applied the changes mentioned [here](https://github.com/paritytech/zombienet-sdk/pull/58#discussion_r1212998241)

crates/provider/src/native.rs
l0r1s: Fields on the struct are private, so the doc comment /// won't be available to end user and will be hidden in the doc, we can replace them with simple comments // and add uppercase on each beginning of the comment to keep it uniform with others comments

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+    tmp_dir:               String,
+    pod_monitor_available: bool,
```
l0r1s, line 43: Naming convention in Rust is is_something

```suggestion
    is_pod_monitor_available: bool,
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+struct FilesystemInMemory {}
+
+#[derive(Debug, Serialize, Clone, PartialEq)]
+struct NativeProvider<T: FileSystem + Debug> {
+    /// Namespace of the client
+    namespace:             String,
+    /// path where configuration relies
+    config_path:           String,
+    // variable that shows if debug is activated
+    debug:                 bool,
```
l0r1s, line 36:
```suggestion
    is_debug:                 bool,
```

crates/provider/src/native.rs
l0r1s: For the reference: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+        config_path: impl Into<String>,
+        tmp_dir: impl Into<String>,
+        filesystem: T,
+    ) -> Self {
+        let tmp_dir = tmp_dir.into();
+
+        Self {
+            namespace: namespace.into(),
+            config_path: config_path.into(),
+            debug: true,
```
l0r1s, line 62:
```suggestion
            is_debug: true,
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+            timeout: 60, // seconds
+            local_magic_file_path: format!("{}/finished.txt", &tmp_dir),
+            remote_dir: format!("{}{}", &tmp_dir, DEFAULT_REMOTE_DIR),
+            data_dir: format!("{}{}", &tmp_dir, DEFAULT_DATA_DIR),
+            command: "bash".into(),
+            tmp_dir,
+            pod_monitor_available: false,
```
l0r1s, line 69:
```suggestion
            is_pod_monitor_available: false,
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+impl<T: FileSystem + Debug> NativeProvider<T> {
+    pub fn new(
+        namespace: impl Into<String>,
+        config_path: impl Into<String>,
+        tmp_dir: impl Into<String>,
+        filesystem: T,
+    ) -> Self {
+        let tmp_dir = tmp_dir.into();
```
l0r1s, line 57:
```suggestion
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+        Self {
+            namespace: namespace.into(),
+            config_path: config_path.into(),
+            debug: true,
+            timeout: 60, // seconds
+            local_magic_file_path: format!("{}/finished.txt", &tmp_dir),
+            remote_dir: format!("{}{}", &tmp_dir, DEFAULT_REMOTE_DIR),
+            data_dir: format!("{}{}", &tmp_dir, DEFAULT_DATA_DIR),
+            command: "bash".into(),
+            tmp_dir,
```
l0r1s, line 68:
```suggestion
            tmp_dir: tmp_dir.into(),
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+            pod_monitor_available: false,
+            filesystem,
+        }
+    }
+}
+
+#[async_trait]
+impl<T: FileSystem + Debug> Provider for NativeProvider<T> {
+    fn create_namespace(&mut self) -> Result<(), Box<dyn Error>> {
+        let name_space_def = NamespaceDef {
```
l0r1s, line 78:
```suggestion
        let namespace_def = NamespaceDef {
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+            api_version: "v1".into(),
+            kind:        "Namespace".into(),
```
l0r1s, line 80: Do you think @pepoviola  we could have enum here and a custom implementation of Deserialize to dump the struct in a clean way to JSON string ?

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+            metadata:    NamespaceMetadata {
+                name:   format!("{}", &self.namespace),
+                labels: None,
+            },
+        };
+
+        let file_path = format!("{}/{}", &self.tmp_dir, "namespace");
```
l0r1s, line 87:
```suggestion
        let file_path = format!("{self.tmp_dir}/namespace");
```

l0r1s, line 87:
```suggestion
        let file_path = format!("{}/namespace", &self.tmp_dir);
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+        let content = serde_json::to_string(&name_space_def)?;
+
+        self.filesystem.write(file_path, content)?;
+        self.filesystem.create_dir(&self.remote_dir)?;
```
l0r1s, line 91: Why do we create a dir after writing the file ? @pepoviola any idea ?

crates/provider/src/native.rs
```diff
@@ -0,0 +1,307 @@
...
+        Ok(())
+    }
+
+    fn setup_cleaner(&self) -> Result<(), Box<dyn Error>> {
+        Ok(())
+    }
```
l0r1s, line 97: If you don't need to implement the method, you can provide a default implementation in the trait: https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations

crates/provider/src/native.rs
wirednkod: This is needed because of the way it is used in the returned `Self`:
```rs
           local_magic_file_path: format!("{}/finished.txt", &tmp_dir), // <---  used here
            remote_dir: format!("{}{}", &tmp_dir, DEFAULT_REMOTE_DIR), // <---  here
            data_dir: format!("{}{}", &tmp_dir, DEFAULT_DATA_DIR), // <---  and here
            command: "bash".into(),
            tmp_dir, // <---  before borrowed here

File: crates/provider/src/native.rs (hunk 53 above)
Comment (pepoviola) [reply crates/provider/src/native.rs:94]: This can be refactored, from this `ts` code

```ts
  async createNamespace(): Promise<void> {
    const namespaceDef = {
      apiVersion: "v1",
      kind: "Namespace",
      metadata: {
        name: this.namespace,
      },
    };

    writeLocalJsonFile(this.tmpDir, "namespace", namespaceDef);
    // Native provider don't have the `namespace` isolation.
    // but we create the `remoteDir` to place files
    await makeDir(this.remoteDir, true);
    return;
  }
  ```
  
  For native we should only do
  
  ```rust
            fn create_namespace(&mut self) -> Result<(), Box<dyn Error>> { 
              self.filesystem.create_dir(&self.remote_dir)?;
              Ok(())
            }
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,336 @@
...
+        mut args: Vec<String>,
+        opts: NativeRunCommandOptions,
+    ) -> Result<RunCommandResponse, Box<dyn Error>> {
+        if let Some(arg) = args.get(0) {
+            if arg == "bash" {
+                args.remove(0);
+            }
+        }
+
+        let output: Output = if cfg!(target_os = "windows") {
```
pepoviola, line 117: I don't think we want to add support for `windows`, at least not now.

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use async_trait::async_trait;
+
+use super::types::{NativeRunCommandOptions, PodDef, RunCommandResponse};
+
+#[async_trait]
+#[allow(non_upper_case_globals)]
+pub trait Provider {
+    fn create_namespace(&mut self) -> Result<(), Box<dyn Error>>;
+    fn setup_cleaner(&self) -> Result<(), Box<dyn Error>>;
```
pepoviola, line 11: Here you can provide the default implementation that @l0r1s  mention [here](https://github.com/paritytech/zombienet-sdk/pull/58/files#r1213485634)

crates/provider/src/native.rs
l0r1s: Alright, I didn't see the ```tmp_dir``` was moved (and not borrowed) using the field init shorthand, an alternative solution would be (just for info) to set something like this on the parameter:

```rust
use core::fmt::Display;

tmp_dir: impl Into<String> + Copy + Display
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+            std_out:   output.stdout,
+            std_err:   Some(output.stderr),
+        })
+    }
+
+    async fn create_resource(&mut self, resourse_def: PodDef) -> Result<(), Box<dyn Error>> {
+        let name: String = resourse_def.metadata.name.clone();
+
+        // This is temporary solution for filling up the process map. To be deleted:
+        self.process_map.insert(
```
pepoviola, line 173: We should use the `Entry` api.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+                std_out:   output.stdout,
+                std_err:   Some(output.stderr),
+            });
+        }
+
+        Ok(RunCommandResponse {
+            exit_code: output.status,
+            std_out:   output.stdout,
+            std_err:   Some(output.stderr),
+        })
```
pepoviola, line 166: I think we can refactor this to make it more simple. I will send you a pr.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+            },
+        );
+        // Delete the code above once spawnFromDef is implemented
+
+        let local_file_path: String = format!("{}/{}.yaml", &self.tmp_dir, name);
+        let content: String = serde_json::to_string(&resourse_def)?;
+
+        self.filesystem
+            .write(&local_file_path, content)
+            .expect("Create source: Failed to write file");
```
pepoviola, line 189: We should return the Error.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+
+        let mut command: String = resourse_def.spec.command.clone();
+        if command.starts_with("bash") {
+            command = command.replace("bash", "");
+        }
+
+        match resourse_def.metadata.labels.zombie_role {
```
pepoviola, line 196: We can refactor this to us `if let` instead this match.

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+            )
+            .expect("Failed to copy file");
+        Ok(())
+    }
+
+    async fn wait_node_ready(&mut self, node_name: String) -> Result<(), Box<dyn Error>> {
+        // check if the process is alive after 1 seconds
+        sleep(Duration::from_millis(1000)).await;
+
+        let process_node_name = self.process_map.get_mut(&node_name).unwrap();
```
pepoviola, line 315:
```suggestion
        let process_node_name = self.process_map.get_mut(&node_name)?;
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+        let result = self
+            .run_command(
+                vec![format!("ps {}", pid)],
+                NativeRunCommandOptions { allow_fail: true },
+            )
+            .await
+            .expect("Failed to run `ps` command");
+
+        if result.exit_code.code().unwrap() > 0 {
+            let lines: String = self.get_node_logs(node_name).await.unwrap();
```
pepoviola, line 329:
```suggestion
            let lines: String = self.get_node_logs(node_name).await?;
```

crates/provider/src/native.rs
```diff
@@ -0,0 +1,601 @@
...
+            .parse::<u32>()
+            .expect("Error while converting 1st time, lines, to u32");
+        let lines_2_output: u32 = String::from_utf8(lines_2.std_out)
+            .unwrap()
+            .parse::<u32>()
+            .expect("Error while converting 2nd time, lines, to u32");
+
+        if lines_2_output > lines_1_output {
+            return Ok(());
+        };
```
pepoviola, line 384: This is just counting that the logs lines are increasing.

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,49 @@
+use std::error::Error;
+
+use async_trait::async_trait;
+
+use super::types::{NativeRunCommandOptions, PodDef, RunCommandResponse};
+
+#[async_trait]
+#[allow(non_upper_case_globals)]
+pub trait Provider {
+    fn create_namespace(&mut self) -> Result<(), Box<dyn Error>>;
```
pepoviola, line 10: this should be async

crates/provider/src/native.rs
wirednkod: Yes... this is what it supposed to do - as taken from the node CLI. There is a comment mentioning:
```rs
 // check log lines grow between 2/6/12 secs
 ```
 but as I put also in comment.. `this looks weird`. Maybe we can discuss it

File: crates/provider/Cargo.toml [hunk 65]
```diff
@@ -6,3 +6,10 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+async-trait = "0.1.68"
+futures = "0.3.28"
+napi = { version="2.12.7", features=["async"]}
+napi-derive = "2.12.5"
+serde = {verson="1.0.163", features = ["derive"]}
+serde_json = "1.0.96"
+tokio = { version = "1.28.2", features = ["process", "macros", "fs", "time"] }
```
l0r1s, line 15: What we could do is [hoist the dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) to the ```workspace's Cargo.toml```, and additionaly only put the version without the minor so the patches are applied automatically at each build:

```rust
[workspace]
members = [
  "crates/support",
  "crates/configuration",
  "crates/orchestrator",
  "crates/provider",
  "crates/test-runner"
]

[workspace.dependencies]
async-trait = "0.1"
futures = "0.3"
napi = "2.12"
napi-derive = "2.12"
serde = "1.0"
serde_json = "1.0"
tokio = "1.28"
```

And then in the ```crate's Cargo.toml``` you can do this:
```suggestion
async-trait = { workspace = true }
futures = { workspace = true }
napi = { workspace = true, features=["async"]}
napi-derive = { workspace = true }
serde = {workspace = true, features = ["derive"]}
serde_json = { workspace = true }
tokio = { workspace = true, features = ["process", "macros", "fs", "time"] }
```

this way, all crates share the same version of dependencies but can get what they want from it.

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
+use std::{error::Error, fs::File};
+
+use crate::{native::FileSystem, shared::types::LocalFile};
+
+#[derive(Debug, PartialEq)]
+pub enum Operation {
+    // DeleteFile { path: String },
+    // DeleteDir,
+    // LinkFile,
```
l0r1s, line 9: Better to remove dead code, when we need it, we will commit it

```suggestion
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[derive(Debug)]
+pub struct MockFilesystem {
+    copy_error:        Option<Box<dyn Error + Send + Sync>>,
+    create_dir_error:  Option<Box<dyn Error + Send + Sync>>,
+    create_file_error: Option<Box<dyn Error + Send + Sync>>,
+    open_file_error:   Option<Box<dyn Error + Send + Sync>>,
+    read_file_error:   Option<Box<dyn Error + Send + Sync>>,
+    write_error:       Option<Box<dyn Error + Send + Sync>>,
+    pub operations:    Vec<Operation>,
```
l0r1s, line 26: Here you can extract a [type alias](https://doc.rust-lang.org/reference/items/type-aliases.html) to make the code cleaner, it can potentially be set in the ```shared.rs``` if reused across files:

```suggestion
type BoxError = Box<dyn Error + Send + Sync>; 

#[derive(Debug)]
pub struct MockFilesystem {
    copy_error:        Option<BoxError>
    create_dir_error:  Option<BoxError>
    create_file_error: Option<BoxError>
    open_file_error:   Option<BoxError>
    read_file_error:   Option<BoxError>
    write_error:       Option<BoxError>,
    pub operations:    Vec<Operation>,
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self {
+            copy_error:        None,
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
```
l0r1s, line 40: When you have all fields as default values of the type (here ```None``` is the default for ```Option<T>``` and ```vec![]``` is the default for ```Vec<T>```, you can directly use ```#[derive(Default)]``` on your struct (here ```MockFilesystem```) and simplify the constructor:

```suggestion
    pub fn new() -> Self {
        Self::default()
    }
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self {
+            copy_error:        None,
+            create_dir_error:  Some(Box::new(error)),
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
```
l0r1s, line 52: Following the previous advice, you can simplify the code of the methods to this:

```suggestion
    fn with_create_dir_error(error: impl Error + Send + Sync + 'static) -> Self {
        Self {
            create_dir_error: Some(Box::new(error)),
            ..Self::default()
        }
    }
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self {
+            copy_error:        None,
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: Some(Box::new(error)),
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
```
l0r1s, line 64: Same here:

```suggestion
    fn with_create_file_error(error: impl Error + Send + Sync + 'static) -> Self {
        Self {
            create_file_error: Some(Box::new(error)),
            ..Self::default()
        }
    }
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self {
+            copy_error:        None,
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   Some(Box::new(error)),
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
```
l0r1s, line 76: Same here:

```suggestion
    fn with_read_file_error(error: impl Error + Send + Sync + 'static) -> Self {
        Self {
            read_file_error: Some(Box::new(error)),
            ..Self::default()
        }
    }
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        Self {
+            copy_error:        Some(Box::new(error)),
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
```
l0r1s, line 88: Same here:

```suggestion
    fn with_copy_error(error: impl Error + Send + Sync + 'static) -> Self {
        Self {
            copy_error: Some(Box::new(error)),
            ..Self::default()
        }
    }
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+            copy_error:        None,
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       Some(Box::new(error)),
+            operations:        vec![],
+        }
+    }
+}
```
l0r1s, line 102: And the last one:

```suggestion
    fn with_write_error(error: impl Error + Send + Sync + 'static) -> Self {
        Self {
            write_error:       Some(Box::new(error)),
            ..Self::default()
        }
    }
}
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
+
+    /// check crate: thisError for easier implementation of errors!
```
l0r1s, line 90: This is an internal comment not a doc:

```suggestion
    // check crate: thisError for easier implementation of errors!
```

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
+use std::error::Error;
+
+use async_trait::async_trait;
+
+use super::types::{NativeRunCommandOptions, PodDef, RunCommandResponse};
+
+#[async_trait]
+#[allow(non_upper_case_globals)]
```
l0r1s, line 8: Not needed

```suggestion
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+            create_dir_error:  None,
+            open_file_error:   None,
+            create_file_error: None,
+            read_file_error:   None,
+            write_error:       None,
+            operations:        vec![],
+        }
+    }
+
+    fn with_create_dir_error(error: impl Error + Send + Sync + 'static) -> Self {
```
l0r1s, line 42: This argument type can be extracted to a generic at the impl level to make the code cleaner:

```rust
impl<E> MockFilesystem
where Err: Error + send + Sync + 'static 
 {
...
}
```

and then be used in all following methods like this:

```suggestion
    fn with_create_dir_error(error: E) -> Self {
```

crates/provider/src/helpers.rs
```diff
@@ -0,0 +1,180 @@
...
+        });
+        Ok(())
+    }
+
+    fn create(&mut self, path: impl Into<String>) -> Result<LocalFile, Box<dyn Error>> {
+        if let Some(err) = self.create_file_error.take() {
+            return Err(err);
+        }
+
+        let p = format!("{}", &path.into());
```
l0r1s, line 136: What is the goal here ?

crates/provider/src/shared/constants.rs
```diff
@@ -0,0 +1,6 @@
+// CONSTANTS
+pub const DEFAULT_REMOTE_DIR: &str = "/cfg";
+pub const DEFAULT_DATA_DIR: &str = "/data";
+pub const LOCALHOST: &str = "127.0.0.1";
```
l0r1s, line 4: Maybe we should use the IpAddr from the stdlib, this is what I use in the configuration crate

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+pub trait Provider {
+    async fn create_namespace(&mut self) -> Result<(), Box<dyn Error>>;
+    fn get_node_ip(&self) -> Result<String, Box<dyn Error>>;
```
l0r1s, line 11: Same here, it can be relevant to use ```IpAddr``` from stdlib to encode our requirement inside the typesystem

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn get_port_mapping(&mut self, port: u16, pod_name: String) -> Result<u16, Box<dyn Error>>;
+    fn get_node_info(&mut self, pod_name: String) -> Result<(String, u16), Box<dyn Error>>;
```
l0r1s, line 13: I would recommend to add a type aliases to make the code clearer:

```rust
type Port = u16;
```
and it can be relevant to use ```IpAddr``` type from std

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+        &self,
+        args: Vec<String>,
+        opts: NativeRunCommandOptions,
+    ) -> Result<RunCommandResponse, Box<dyn Error>>;
+    async fn run_script(
+        &mut self,
+        identifier: String,
+        script_path: String,
+        args: Vec<String>,
+    ) -> Result<RunCommandResponse, Box<dyn Error>>;
```
l0r1s, line 24: Why does ```run_script``` returns a ```RunCommandResponse``` and not a ```RunScriptResponse``` ?

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn copy_file_from_pod(
+        &mut self,
+        pod_file_path: String,
+        local_file_path: String,
```
l0r1s, line 28: Can be relevant to use ```PathBuf``` type from crate ```path``` to encode our requirement in the typesystem

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+    ) -> Result<(), Box<dyn Error>>;
+    async fn create_resource(&mut self, resource_def: PodDef) -> Result<(), Box<dyn Error>>;
+    async fn wait_node_ready(&mut self, node_name: String) -> Result<(), Box<dyn Error>>;
+    async fn get_node_logs(&mut self, node_name: String) -> Result<String, Box<dyn Error>>;
+    async fn dump_logs(&mut self, path: String, pod_name: String) -> Result<(), Box<dyn Error>>;
```
l0r1s, line 33: ```path::PathBuf```

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn get_pause_args(&mut self, name: String) -> Vec<String>;
+    fn get_resume_args(&mut self, name: String) -> Vec<String>;
+    async fn validate_access(&mut self) -> Result<bool, Box<dyn Error>>;
```
l0r1s, line 36: When I see ```validate_access```, it's not straightforward to understand what kind of access we validate ?

crates/provider/src/shared/provider.rs
```diff
@@ -0,0 +1,63 @@
...
+        Ok(())
+    }
+    fn create_pod_monitor() -> Result<(), Box<dyn Error>> {
+        Ok(())
+    }
+    fn setup_cleaner() -> Result<(), Box<dyn Error>> {
+        Ok(())
+    }
+    fn upsert_cron_job() -> Result<(), Box<dyn Error>> {
+        todo!();
```
l0r1s, line 58: ```todo!()``` is ```unimplemented!()``` wrapped so it will panic when called

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum ZombieRole {
+    Temp,
+    Node,
+    BootNode,
+    Collator,
+    CumulusCollator,
+    Authority,
+    FullNode,
```
l0r1s, line 16: Are all these types used @pepoviola because from what I've seen in the typescript version, there is an overlap between ```label``` and ```role``` in the same enum but not all variant are really used for ```role```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,185 @@
...
+    IfNotPresent,
+    Never,
+    Always,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct FileMap {
+    local_file_path:  PathBuf,
+    remote_file_path: PathBuf,
+    unique:           bool,
```
l0r1s, line 38: What does ```unique``` means here ?

crates/provider/src/shared/types.rs
l0r1s: I think it's better to have a String, that way, when we use a created ```RunCommandResponse.std_err``` we don't need to use a conversion first to use it because it was done at instanciation

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,185 @@
...
+            exit_code: ExitStatus::from_raw(0),
+            std_out:   vec![],
+            std_err:   None,
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct NativeRunCommandOptions {
+    pub allow_fail: bool,
```
l0r1s, line 60: Following [rust convention](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#predicates), we should prefix boolean with ```is_```:

```suggestion
    pub is_failure_allowed: bool,
```

crates/provider/src/shared/types.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub env:       ProcessEnvironment,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct PodDef {
+    pub metadata: PodMetadata,
+    pub spec:     PodSpec,
+}
+
+type ProcessEnvironment = HashMap<String, String>;
```
l0r1s, line 113: I don't know if it's possible to use ```HashMap::from([("foo", "bar"), ("bar"baz")]```, with this type ?

If not, it will force us to first create a hash map and insert k/v individually, which is not practical, maybe we could replace this with:

```rust
struct EnvVar {
   name: String,
   value: String
}

impl From<(&str, &str)> for EnvVar {
    fn into(value: (&str, &str) -> Self {
        Self { name: value.0.into(), value: value.1.into() }    
    }
}
```

this way we can use this directly with ```let environment: Vec<EnvVar> = vec![("foo", "bar").into(), ("bar", "baz).into()];```

crates/provider/src/shared/types.rs
l0r1s: @pepoviola @wirednkod I think this is time to revisit options ?

l0r1s (changes requested):
Some typefu and questions we need to answer imho but good works ! :muscle:

crates/provider/src/shared/types.rs
pepoviola: `unique` is a flag to identify if the file should be cached in the fileserver or not.

crates/provider/src/native.rs
```diff
@@ -219,6 +219,19 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+    fn copy_file_from_pod(
+        &mut self,
+        pod_file_path: String,
+        local_file_path: String,
+    ) -> Result<(), Box<dyn Error>> {
+        // TODO: log::debug!(format!("cp {} {}", pod_file_path, local_file_path));
+
+        self.filesystem
+            .copy(&pod_file_path, &local_file_path)
+            .expect("Failed to copy file");
```
pepoviola, line 231: we should not use `expect` and just return the error.

crates/provider/src/shared/provider.rs
pepoviola: Yes,  or we can just rename/abstract to use in both places. Since this type only represent the result of some execution (bash/js).

crates/provider/src/shared/types.rs
l0r1s: Maybe we could replace it with something more explicit like ```is_cached``` ?

pepoviola: We can use `is_unique`, since here if is unique we don't need to cache the file.

crates/support/src/net.rs
```diff
@@ -0,0 +1,10 @@
+use std::io::Cursor;
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error + Send + Sync>>;
+
+pub async fn download_file(url: String, dest: String) -> Result<()> {
```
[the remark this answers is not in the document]
wirednkod: What is passed is the string of the path (`format!(...)`). I think that it does not mean that the calling scope is loosing access to that path

crates/provider/src/shared/provider.rs
wirednkod: @l0r1s `RunCommandResponse` was created initially for the response of `run_command` and was reused afterwards; If renamed to `RunScriptResponse`, then the `run_command` will have a `RunScriptResponse` as a response, which is the similar "issue" revered 😄

crates/provider/src/shared/provider.rs
wirednkod: Actually this is just running the `--help` command so I will rename accordingly to `get_help_info`

rustfmt.toml
```diff
@@ -10,8 +10,8 @@ use_field_init_shorthand = true
 use_try_shorthand = true
 
 # alignment
-struct_field_align_threshold = 100
-enum_discrim_align_threshold = 100
+struct_field_align_threshold = 0
+enum_discrim_align_threshold = 0
```
l0r1s, line 14: I think you can remove them directly, the default is 0

```suggestion
```

crates/support/Cargo.toml
```diff
@@ -6,3 +6,8 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+thiserror = { workspace = true }
+async-trait = { workspace = true }
+futures = { workspace = true }
+reqwest = "0.11.18"
```
l0r1s, line 12: Maybe put this one too in the workspace root ? And allow patch using "0.11"

crates/provider/src/lib.rs
```diff
@@ -1,14 +1,93 @@
...
+    async fn wait_node_ready(&mut self, node_name: String) -> Result<(), ProviderError>;
+    async fn get_node_logs(&mut self, node_name: String) -> Result<String, ProviderError>;
+    async fn dump_logs(&mut self, path: String, pod_name: String) -> Result<(), ProviderError>;
+    fn get_pause_args(&mut self, name: String) -> Vec<String>;
+    fn get_resume_args(&mut self, name: String) -> Vec<String>;
+    async fn restart_node(&mut self, name: String, timeout: u64) -> Result<bool, ProviderError>;
+    async fn get_help_info(&mut self) -> Result<(), ProviderError>;
+    async fn destroy_namespace(&mut self) -> Result<(), ProviderError>;
+    async fn get_logs_command(&mut self, name: String) -> Result<String, ProviderError>;
+    // TODO: need to implement
```
l0r1s, line 60: Needed ?

l0r1s (approved):

pepoviola (approved):

PR #59, opened 2023-05-25
explore builders and how to use them

pepoviola:
[DRAFT], just to open the discussion about the API:

-  The `NetworkConfig` now should be built using the `NetworkConfigBuilder` (with type state to ensure `.with_relaychain` is called ones).
- Add same `Defaults` we use now in the `ts` version.
- Add `choice` to methods like `being_validator` since the default is `true` and we want to allow to set to false. We should revisit the naming here.
- Remove Settings/ options that we should revisit.
- Add `thiserror` to handle `ConfigErrors`
- Add serde for `dump` the config to json.
- Lots of `TODO` to visit and add test :)

### DIFF
--- a/crates/configuration/src/main.rs
+++ b/crates/configuration/src/main.rs
@@ -1,0 +1,12 @@
+use configuration::{ConfigError, NetworkConfigBuilder};
+
+pub fn main() -> Result<(), ConfigError> {
+    let network_config = NetworkConfigBuilder::new()
+        .with_global_settings(|g| g)
+        .with_relaychain(|r| r)
+        .with_parachain(|para| para)
+        .build()?;
+
+    println!("{}", network_config.dump()?);
+    Ok(())
+}
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -22,7 +22,7 @@ pub struct GlobalSettings {
 
     /// External bootnode address.
     /// [TODO]: is it a default overriden by node config, maybe an option ?
-    bootnode_address: Vec<MultiAddress>,
+    bootnodes_addresses: Vec<MultiAddress>,
 
     /// Global spawn timeout in seconds.
     network_spawn_timeout: TimeoutInSecs,
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -20,7 +20,7 @@ pub struct GlobalSettings {
     /// Whether we should spawn a dedicated bootnode for each chain.
     // spawn_bootnode: bool,
 
-    /// External bootnode address.
+    /// External bootnodes addresses.
     /// [TODO]: is it a default overriden by node config, maybe an option ?
     bootnodes_addresses: Vec<MultiAddress>,
 
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -1,22 +1,34 @@
+use core::marker::PhantomData;
+
+use serde::Serialize;
+
 use crate::{
-    shared::types::{Duration, IpAddress, MultiAddress},
+    errors::ConfigError,
+    shared::types::{IpAddress, MultiAddress, ParaId, TimeoutInSecs},
     HrmpChannelConfig, ParachainConfig, RelaychainConfig,
 };
 
+#[derive(Default, Clone)]
+pub struct NoRelayChain;
+#[derive(Default, Clone)]
+pub struct WithRelayChain;
+
 /// Global settings applied to an entire network.
+#[derive(Debug, Serialize)]
 pub struct GlobalSettings {
+    /// [TODO]: This feature is not well defined, which binary/image should use to spawn the bootnode?
     /// Whether we should spawn a dedicated bootnode for each chain.
-    spawn_bootnode: bool,
+    // spawn_bootnode: bool,
 
     /// External bootnode address.
     /// [TODO]: is it a default overriden by node config, maybe an option ?
-    bootnode_address: MultiAddress,
+    bootnode_address: Vec<MultiAddress>,
 
     /// Global spawn timeout in seconds.
-    network_spawn_timeout: Duration,
+    network_spawn_timeout: TimeoutInSecs,
 
     /// Individual node spawn timeout.
-    node_spawn_timeout: Duration,
+    node_spawn_timeout: TimeoutInSecs,
 
     /// Local IP used to expose local services (including RPC, metrics and monitoring).
     local_ip: Option<IpAddress>,
@@ -24,12 +36,18 @@ pub struct GlobalSettings {
 
 impl Default for GlobalSettings {
     fn default() -> Self {
-        // [TODO]: define the default value for global settings
-        todo!()
+        Self {
+            // spawn_bootnode: false,
+            bootnode_address:      vec![],
+            network_spawn_timeout: 1000,
+            node_spawn_timeout:    300,
+            local_ip:              None,
+        }
     }
 }
 
 /// A network configuration, composed of a relaychain, parachains and HRMP channels.
+#[derive(Debug, Serialize)]
 pub struct NetworkConfig {
     /// The global settings applied to the network.
     global_settings: GlobalSettings,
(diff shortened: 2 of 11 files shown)


crates/configuration/src/main.rs
```diff
@@ -0,0 +1,12 @@
+use configuration::{ConfigError, NetworkConfigBuilder};
+
+pub fn main() -> Result<(), ConfigError> {
+    let network_config = NetworkConfigBuilder::new()
+        .with_global_settings(|g| g)
+        .with_relaychain(|r| r)
```
pepoviola, line 6: I think we can add another method here, `with_default_relaychain` that just change the `type state`, allowing users to use the default `rococo-local` (w/o any customization).

crates/configuration/src/shared/node.rs
```diff
@@ -99,23 +100,23 @@ impl NodeConfig {
...
+    pub fn being_invulnerable(self, choice: bool) -> Self {
         Self {
-            is_invulnerable: true,
+            is_invulnerable: choice,
             ..self
         }
     }
 
-    pub fn being_bootnode(self) -> Self {
+    pub fn being_bootnode(self, choice: bool) -> Self {
```
pepoviola, line 117: naming is hard :)

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
 pub struct GlobalSettings {
+    /// [TODO]: This feature is not well defined, which binary/image should use to spawn the bootnode?
     /// Whether we should spawn a dedicated bootnode for each chain.
-    spawn_bootnode: bool,
+    // spawn_bootnode: bool,
 
     /// External bootnode address.
     /// [TODO]: is it a default overriden by node config, maybe an option ?
-    bootnode_address: MultiAddress,
+    bootnode_address: Vec<MultiAddress>,
```
l0r1s, line 25:
```suggestion
    bootnodes_addresses: Vec<MultiAddress>,
```

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
+
 /// Global settings applied to an entire network.
+#[derive(Debug, Serialize)]
 pub struct GlobalSettings {
+    /// [TODO]: This feature is not well defined, which binary/image should use to spawn the bootnode?
     /// Whether we should spawn a dedicated bootnode for each chain.
-    spawn_bootnode: bool,
+    // spawn_bootnode: bool,
 
     /// External bootnode address.
```
l0r1s, line 23:
```suggestion
    /// External bootnodes addresses.
```

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
 
+#[derive(Default, Clone)]
+pub struct NoRelayChain;
+#[derive(Default, Clone)]
+pub struct WithRelayChain;
+
 /// Global settings applied to an entire network.
+#[derive(Debug, Serialize)]
 pub struct GlobalSettings {
+    /// [TODO]: This feature is not well defined, which binary/image should use to spawn the bootnode?
```
l0r1s, line 19: I would say the default image/binary on a parachain or relaychain, it makes this property dependant on the default_image/default_command on underlying type in validation ?

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
+
+    /// Parachains configurations.
+    parachains: Vec<ParachainConfig>,
+
+    /// HRMP channels configurations.
+    hrmp_channels: Vec<HrmpChannelConfig>,
 
-    // [TODO]: what does it represents ?
-    seed: String,
+    typestate: PhantomData<R>,
```
l0r1s, line 80: :+1:

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
+    pub fn new() -> Self {
+        NetworkConfigBuilder::default()
     }
 }
 
-impl NetworkConfig {
-    pub fn new() -> NetworkConfig {
-        Self::default()
+impl NetworkConfigBuilder<WithRelayChain> {
+    pub fn build(self) -> Result<NetworkConfig, ConfigError> {
```
l0r1s, line 90: I think we should returns a Vec<ConfigError> with specific path of the failing key to guide the user.

For example having something like this:

Error: Cause of the error in "network.relaychain.default_command"

wdyt ?

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
+
+    pub(crate) fn parachains(&self) -> &Vec<ParachainConfig> {
+        &self.parachains
+    }
+
+    pub(crate) fn hrmp_channels(&self) -> &Vec<HrmpChannelConfig> {
+        &self.hrmp_channels
+    }
+
+    //[TODO]: skill serializing empty vec or None
```
l0r1s, line 167: I don't understand the comment here

crates/configuration/src/shared/node.rs
```diff
@@ -27,7 +28,7 @@ pub struct NodeConfig {
     image: Option<String>,
 
     /// Command to run the node. Override the default.
-    command: Option<String>,
+    command: String,
```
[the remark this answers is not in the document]
l0r1s: Yes, today we don't have the default_command and the default_image on the Parachain, it makes sense do add them to the struct and use the type state on the Node

crates/configuration/src/shared/node.rs
l0r1s: Yes, here I chose being_bootnode because with_bootnode didn't make sense and is_bootnode is the name of the getter, if you find better naming, I'm open !

l0r1s:
> [DRAFT], just to open the discussion about the API:
> 
>     * The `NetworkConfig` now should be built using the `NetworkConfigBuilder` (with type state to ensure `.with_relaychain` is called ones).
> 
>     * Add same `Defaults` we use now in the `ts` version.
> 
>     * Add `choice` to methods like `being_validator` since the default is `true` and we want to allow to set to false. We should revisit the naming here.
> 
>     * Remove Settings/ options that we should revisit.
> 
>     * Add `thiserror` to handle `ConfigErrors`
> 
>     * Add serde for `dump` the config to json.
> 
>     * Lots of `TODO` to visit and add test :)

Thank you, good improvements ! Added some comments :+1:

crates/configuration/src/network.rs
pepoviola: Yes, we can use that approach and add the docs to explain the expected behavior. At the moment, in the current version, we take the image from the `first` node in the relaychain. I think is ok, but we should add a nice documentation.
Thanks!

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
+    // spawn_bootnode: bool,
 
     /// External bootnode address.
     /// [TODO]: is it a default overriden by node config, maybe an option ?
-    bootnode_address: MultiAddress,
+    bootnode_address: Vec<MultiAddress>,
 
     /// Global spawn timeout in seconds.
-    network_spawn_timeout: Duration,
+    network_spawn_timeout: TimeoutInSecs,
```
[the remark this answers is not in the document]
pepoviola: `Timeout` will be always in seconds, we can use just `Timeout` and add in the docs that is always in seconds. Or we can move to `Duration` type and allow users to specify as they want.

crates/configuration/src/network.rs
pepoviola: Sounds good, a Vec of error is great and we can keep in mind to implement in a follow-up pr.

crates/configuration/src/network.rs
pepoviola: We need to mark the fields to `skip if...` in serde to not not include `null`s in the json output (some empty vecs can also be skipped)

```json
{
  "global_settings": {
    "bootnode_address": [],
    "network_spawn_timeout": 1000,
    "node_spawn_timeout": 300,
    "local_ip": null
  },
  "relaychain": {
    "default_command": "polkadot",
    "default_image": "parity/polkadot:latest",
    "default_resources": null,
    "default_db_snapshot": null,
    "chain": "rococo-local",
    "chain_spec_path": null,
    "default_args": [],
    "random_nominators_count": null,
    "max_nominations": null,
    "nodes": []
  },
  "parachains": [
    {
      "id": 100,
      "chain": null,
      "registration_strategy": null,
      "initial_balance": 1000000000,
      "genesis_wasm_path": null,
      "genesis_wasm_generator": null,
      "genesis_state_path": null,
      "genesis_state_generator": null,
      "chain_spec_path": null,
      "is_cumulus_based": true,
      "bootnodes_addresses": [],
      "collators": []
    }
  ],
  "hrmp_channels": []
}
```

crates/configuration/src/network.rs
wirednkod: I think it is clearer to the user (better DX) to have an extra param in the configuration called "create_bootnode"; If used then the node with that param will be used for creating the bootnodes; If not then by default the `first` node in the relaychain

crates/configuration/src/network.rs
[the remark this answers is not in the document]
wirednkod: FWIW the less docs the better - I mean that the code should be straightforward - As @pepoviola mentions `Timeout` is (and should be IMO) always in seconds. Having a `Duration` will help us to allow something different. I agree

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
 
     /// Individual node spawn timeout.
-    node_spawn_timeout: Duration,
+    node_spawn_timeout: TimeoutInSecs,
 
     /// Local IP used to expose local services (including RPC, metrics and monitoring).
     local_ip: Option<IpAddress>,
 }
 
 impl Default for GlobalSettings {
```
wirednkod, line 37:
```suggestion
impl Default for Settings {
```

crates/configuration/src/network.rs
```diff
@@ -1,35 +1,53 @@
...
+            local_ip:              None,
+        }
     }
 }
 
 /// A network configuration, composed of a relaychain, parachains and HRMP channels.
+#[derive(Debug, Serialize)]
 pub struct NetworkConfig {
     /// The global settings applied to the network.
     global_settings: GlobalSettings,
```
wirednkod, line 53:
```suggestion
    settings: Settings,
```

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
 
     /// HRMP channels configurations.
     hrmp_channels: Vec<HrmpChannelConfig>,
+}
+
+/// A network configuration, composed of a relaychain, parachains and HRMP channels.
+#[derive(Debug, Default)]
+pub struct NetworkConfigBuilder<R> {
+    /// The global settings applied to the network.
+    global_settings: GlobalSettings,
```
wirednkod, line 69:
```suggestion
    settings: Settings,
```

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
+        })
     }
+}
 
-    pub fn with_global_settings(self, f: fn(GlobalSettings) -> GlobalSettings) -> Self {
-        Self {
+impl<R> NetworkConfigBuilder<R> {
+    pub fn with_global_settings(
+        self,
+        f: fn(GlobalSettings) -> GlobalSettings,
```
wirednkod, line 104:
```suggestion
        f: fn(Settings) -> Settings,
```

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
+    ) -> NetworkConfigBuilder<R> {
+        NetworkConfigBuilder {
             global_settings: f(self.global_settings),
```
wirednkod, line 107:
```suggestion
            settings: f(self.settings),
```

crates/configuration/src/network.rs
```diff
@@ -42,48 +60,114 @@ pub struct NetworkConfig {
...
-            relaychain: f(self.relaychain),
-            ..self
+impl NetworkConfigBuilder<NoRelayChain> {
+    pub fn with_relaychain(
+        self,
+        f: fn(RelaychainConfig) -> RelaychainConfig,
+    ) -> NetworkConfigBuilder<WithRelayChain> {
+        NetworkConfigBuilder {
+            relaychain:      f(RelaychainConfig::default()),
+            global_settings: self.global_settings,
```
wirednkod, line 120:
```suggestion
            settings: self.settings,
```

crates/configuration/src/shared/node.rs
wirednkod: I think simply `bootnode` is enough for this..

PR #69, opened 2023-06-05
Simplify the node builder to allow default_command

l0r1s:
We want the user to be able to set a default command on the relaychain and to override it, to make it type safe, using a Either type with both possible options and a further step of validation to ensure the user defined a command on the node, here a code example with a schema showing what we want to achieve:

sdk_either_initial_default_command

### DIFF
--- a/crates/configuration/src/relaychain.rs
+++ b/crates/configuration/src/relaychain.rs
@@ -87,8 +87,7 @@ states! {
     Initial,
     WithChain,
     WithAtLeastOneNode,
-    WithDefaultCommand,
-    WithDefaultCommandAndAtLeastOneNode
+    WithDefaultCommand
 }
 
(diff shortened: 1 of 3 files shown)


crates/configuration/src/relaychain.rs
```diff
@@ -343,4 +334,7 @@ mod tests {
         assert_eq!(relaychain_config.random_minators_count().unwrap(), 42);
         assert_eq!(relaychain_config.max_nominations().unwrap(), 5);
     }
+
+    #[test]
+    fn debug() {}
```
pepoviola, line 339:
```suggestion
```
remove debug.

l0r1s:
Either doesn't bring anything regarding type checking here so removed as discussed

wirednkod (approved):

PR #78, opened 2023-06-20
Implement pause, resume, restart

wirednkod:
- Fix warnings;
- Implement pause, resume, restart

### DIFF
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -16,4 +16,3 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = {workspace = true}
 tokio = { workspace = true, features = ["process", "macros", "fs", "time", "rt"] }
 thiserror = {workspace = true}
-sysinfo = "0.29.2"
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -16,3 +16,4 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = {workspace = true}
 tokio = { workspace = true, features = ["process", "macros", "fs", "time", "rt"] }
 thiserror = {workspace = true}
+sysinfo = "0.29.2"
--- a/crates/provider/src/native.rs
+++ b/crates/provider/src/native.rs
@@ -289,17 +289,17 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     // TODO: Add test
     async fn pause(&self, node_name: &str) -> Result<(), ProviderError> {
-        let process_id: Result<u32, ProviderError> = match self.process_map.get(node_name) {
-            Some(process) => Ok(process.pid),
-            None => Err(ProviderError::MissingNodeInfo(
+        let process = self
+            .process_map
+            .get(node_name)
+            .ok_or(ProviderError::MissingNodeInfo(
                 node_name.to_owned(),
                 "process".into(),
-            )),
-        };
+            ))?;
 
         let _ = self
             .run_command(
-                vec![format!("kill -TSTP {}", process_id.unwrap())],
+                vec![format!("kill -STOP {}", process.pid)],
                 NativeRunCommandOptions {
                     is_failure_allowed: true,
                 },
@@ -310,17 +310,17 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     // TODO: Add test
     async fn resume(&self, node_name: &str) -> Result<(), ProviderError> {
-        let process_id: Result<u32, ProviderError> = match self.process_map.get(node_name) {
-            Some(process) => Ok(process.pid),
-            None => Err(ProviderError::MissingNodeInfo(
+        let process = self
+            .process_map
+            .get(node_name)
+            .ok_or(ProviderError::MissingNodeInfo(
                 node_name.to_owned(),
                 "process".into(),
-            )),
-        };
+            ))?;
 
         let _ = self
             .run_command(
-                vec![format!("kill -CONT {}", process_id.unwrap())],
+                vec![format!("kill -CONT {}", process.pid)],
                 NativeRunCommandOptions {
                     is_failure_allowed: true,
                 },
@@ -337,7 +337,7 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
             .ok_or(ProviderError::MissingNodeInfo(
                 node_name.to_owned(),
                 "process".into(),
-            ));
+            ))?;
         self.run_command(
             [format!("kill -9 {:?}", pid)].to_vec(),
             NativeRunCommandOptions {
--- a/crates/provider/src/native.rs
+++ b/crates/provider/src/native.rs
@@ -1,9 +1,6 @@
 use std::{
     self,
-    collections::{
-        hash_map::Entry::{Occupied, Vacant},
-        HashMap,
-    },
+    collections::HashMap,
     fmt::Debug,
     net::IpAddr,
     path::{Path, PathBuf},
@@ -11,20 +8,15 @@ use std::{
 
 use async_trait::async_trait;
 use serde::Serialize;
-use support::{fs::FileSystem, net::download_file};
-use tokio::{
-    process::Command,
-    time::{sleep, Duration},
-};
+use support::fs::FileSystem;
+use tokio::process::Command;
 
 use super::Provider;
 use crate::{
     errors::ProviderError,
     shared::{
         constants::{DEFAULT_DATA_DIR, DEFAULT_REMOTE_DIR, LOCALHOST, P2P_PORT},
-        types::{
-            FileMap, NativeRunCommandOptions, Node, Port, Process, RunCommandResponse, ZombieRole,
-        },
+        types::{FileMap, NativeRunCommandOptions, Node, Port, Process, RunCommandResponse},
     },
 };
 #[derive(Debug, Serialize, Clone, PartialEq)]
@@ -61,7 +53,7 @@ impl<T: FileSystem + Send + Sync> NativeProvider<T> {
         tmp_dir: impl Into<String>,
         filesystem: T,
     ) -> Self {
-        let e = std::io::Error::new(std::io::ErrorKind::BrokenPipe, "error");
+        let _e = std::io::Error::new(std::io::ErrorKind::BrokenPipe, "error");
 
         let tmp_dir: String = tmp_dir.into();
         let process_map: HashMap<String, Process> = HashMap::new();
@@ -139,10 +131,10 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     async fn spawn_node(
         &mut self,
-        node: Node,
-        files_inject: Vec<FileMap>,
-        keystore: &str,
-        db_snapshot: &str,
+        _node: Node,
+        _files_inject: Vec<FileMap>,
+        _keystore: &str,
+        _db_snapshot: &str,
     ) -> Result<(), ProviderError> {
         // TODO: We should implement the logic to go from the `Node` (nodeSpec)
         // to the running node, since we will no expose anymore the underline `Def`.
@@ -153,9 +145,9 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     async fn spawn_temp(
         &self,
-        node: Node,
-        files_inject: Vec<FileMap>,
-        files_get: Vec<FileMap>,
+        _node: Node,
+        _files_inject: Vec<FileMap>,
+        _files_get: Vec<FileMap>,
     ) -> Result<(), ProviderError> {
         // TODO: We should implement the logic to go from the `Node` (nodeSpec)
         // to the running node, since we will no expose anymore the underline `Def`.
@@ -296,17 +288,17 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
         Ok(format!("tail -f {}/{}.log", self.tmp_dir, name))
     }
 
-    async fn pause(&self, node_name: &str) -> Result<(), ProviderError> {
+    async fn pause(&self, _node_name: &str) -> Result<(), ProviderError> {
         // TODO: impl using run_command
         Ok(())
     }
 
-    async fn resume(&self, node_name: &str) -> Result<(), ProviderError> {
+    async fn resume(&self, _node_name: &str) -> Result<(), ProviderError> {
         // TODO: impl using run_command
         Ok(())
     }
 
-    async fn restart(&mut self, node_name: &str, after_secs: u16) -> Result<bool, ProviderError> {
+    async fn restart(&mut self, _node_name: &str, _after_secs: u16) -> Result<bool, ProviderError> {
         // TODO: impl using run_command
         // use &mut here since we mostly need to update the the pid/maps
         Ok(true)
@@ -317,7 +309,7 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
         Ok((LOCALHOST, host_port))
     }
 
-    async fn get_node_ip(&self, node_name: &str) -> Result<IpAddr, ProviderError> {
+    async fn get_node_ip(&self, _node_name: &str) -> Result<IpAddr, ProviderError> {
         Ok(LOCALHOST)
     }
 
@@ -347,7 +339,6 @@ mod tests {
     use support::fs::mock::{MockError, MockFilesystem, Operation};
 
     use super::*;
-    use crate::shared::types::{PodLabels, PodMetadata, PodSpec};
 
     #[test]
     fn new_native_provider() {
--- a/crates/provider/src/native.rs
+++ b/crates/provider/src/native.rs
@@ -357,12 +357,12 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
                     "process".into(),
                 ))?;
 
-        let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
+        // let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
 
         let child_process: Child = Command::new(self.command.clone())
             .arg("-c")
             .arg(process.command.clone())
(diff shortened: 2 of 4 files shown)


crates/provider/Cargo.toml
```diff
@@ -16,3 +16,4 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = {workspace = true}
 tokio = { workspace = true, features = ["process", "macros", "fs", "time", "rt"] }
 thiserror = {workspace = true}
+sysinfo = "0.29.2"
```
pepoviola, line 19:
```suggestion
```
We don't need this dep now.

crates/provider/src/native.rs
```diff
@@ -61,7 +54,7 @@ impl<T: FileSystem + Send + Sync> NativeProvider<T> {
         tmp_dir: impl Into<String>,
         filesystem: T,
     ) -> Self {
-        let e = std::io::Error::new(std::io::ErrorKind::BrokenPipe, "error");
+        let _e = std::io::Error::new(std::io::ErrorKind::BrokenPipe, "error");
```
pepoviola, line 57:
```suggestion
```
don't used, could be something that was a test or something?

crates/provider/src/native.rs
```diff
@@ -296,19 +289,83 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            Some(process) => Ok(process.pid),
+            None => Err(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            )),
+        };
+
+        let _ = self
+            .run_command(
+                [format!("kill -TSTP {}", process_id.unwrap())].to_vec(),
```
pepoviola, line 304:
```suggestion
                vec![format!("kill -STOP {}", process_id.unwrap())],
```

crates/provider/src/native.rs
wirednkod: Could be. Since it was your wip I just added the `_` so that I wont remove code that you may need

crates/provider/src/native.rs
```diff
@@ -296,19 +289,83 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            Some(process) => Ok(process.pid),
+            None => Err(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            )),
+        };
+
+        let _ = self
+            .run_command(
+                [format!("kill -CONT {}", process_id.unwrap())].to_vec(),
```
pepoviola, line 325:
```suggestion
                vec![format!("kill -CONT {}", process_id.unwrap())],
```

crates/provider/src/native.rs
```diff
@@ -296,19 +289,83 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+
+        let res = Command::new(command).arg("-c").envs(&mapped_env).spawn();
+
+        let restarted_process = match res {
+            Ok(process) => process,
+            Err(e) => return Err(ProviderError::ErrorSpawningNode(e.to_string())),
+        };
+
+        process.pid = restarted_process.id().unwrap();
+
```
pepoviola, line 368: This will not work  as expected, since the `clone` will create a new `Process` and you are then shadowing as mut but never changing the underline process in the map. Look this [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f08d3da94aafa1d164c5aa4521dd0f6b)

You need to use `[get_mut](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get_mut)` in order to get a mutable ref. Ping me and we can touch base on this one.

crates/provider/src/native.rs
```diff
@@ -296,19 +287,85 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            Some(process) => Ok(process.pid),
+            None => Err(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            )),
+        };
+
+        let _ = self
+            .run_command(
+                vec![format!("kill -CONT {}", process_id.unwrap())],
```
pepoviola, line 323: `unwrap` will panic here if the `node_name` is not in the process_map. You can use `ok_or` and bubble up the error like this

```rust
        let process = self.process_map.get(node_name).ok_or(ProviderError::MissingNodeInfo(
            node_name.to_owned(),
            "process".into(),
        ))?;

        let _ = self
            .run_command(
                vec![format!("kill -CONT {}", process.pid)],
                NativeRunCommandOptions {
                    is_failure_allowed: true,
                },
            )
            .await?;
        Ok(())
  ```

File: crates/provider/src/native.rs [hunk 7]
```diff
@@ -296,19 +287,85 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
-        // use &mut here since we mostly need to update the the pid/maps
+    // TODO: Add test
+    async fn restart(&mut self, node_name: &str, _after_secs: u16) -> Result<bool, ProviderError> {
+        let pid = self
+            .process_map
+            .get(node_name)
+            .ok_or(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            ));
```
pepoviola, line 340:
```suggestion
            ))?;
```
missing `?` to get the Ok value / return the Err

crates/provider/src/native.rs
```diff
@@ -296,19 +287,85 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+                      "process".into(),));
+      };
+
+        let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
+
+        let res = Command::new(self.command.clone())
+            .arg("-c")
+            .envs(&mapped_env)
+            .spawn();
+
```
pepoviola, line 361: This will spawn an `empty` bash process, we should use the `command` from the `process_map`

Something like
```rust
        let res = Command::new(self.command.clone())
            .arg("-c")
            .arg(process.command.clone())
            .envs(&mapped_env)
            .spawn();
```

wirednkod: Why not this?
```rs
        let res = Command::new(process.command.clone())
            .arg("-c")
            .envs(&mapped_env)
            .spawn();
```

crates/provider/src/native.rs
```diff
@@ -296,19 +287,80 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
-        // use &mut here since we mostly need to update the the pid/maps
+    // TODO: Add test
+    async fn restart(&mut self, node_name: &str, _after_secs: u16) -> Result<bool, ProviderError> {
+        let pid = self
+            .process_map
+            .get(node_name)
+            .ok_or(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            ))?;
```
pepoviola, line 340: Here you are getting the process, not the `pid`.
```suggestion
        let process = self
            .process_map
            .get(node_name)
            .ok_or(ProviderError::MissingNodeInfo(
                node_name.to_owned(),
                "process".into(),
            ))?;
```

crates/provider/src/native.rs
```diff
@@ -296,19 +287,80 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+        self.run_command(
+            [format!("kill -9 {:?}", pid)].to_vec(),
```
pepoviola, line 342: Since we get the `process` above.
```suggestion
            vec![format!("kill -9 {:?}", process.pid)],
```

crates/provider/src/native.rs
```diff
@@ -296,19 +287,80 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+                is_failure_allowed: true,
+            },
+        )
+        .await?;
+
+        let Some(process) = self.process_map.get_mut(node_name) else {
+          return Err(ProviderError::MissingNodeInfo(
+                      node_name.to_owned(),
+                      "process".into(),));
+      };
```
pepoviola, line 353: We can use the same approach with ok_or

```suggestion
        let process = self
            .process_map
            .get_mut(node_name)
            .ok_or(ProviderError::MissingNodeInfo(
                node_name.to_owned(),
                "process".into(),
            ))?;
```

crates/provider/src/native.rs
```diff
@@ -296,19 +287,80 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+
+        let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
+
+        Command::new(self.command.clone())
+            .arg("-c")
+            .arg(process.command.clone())
+            .envs(&mapped_env)
+            .spawn()
+            .map_err(|e| ProviderError::ErrorSpawningNode(e.to_string()))?;
```
pepoviola, line 362: We are not updating the `pid` in the `process_map`, we should get the child process pid and update the map.

crates/provider/src/native.rs
```diff
@@ -139,10 +130,10 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     async fn spawn_node(
         &mut self,
-        node: Node,
-        files_inject: Vec<FileMap>,
-        keystore: &str,
-        db_snapshot: &str,
+        _node: Node,
+        _files_inject: Vec<FileMap>,
```
l0r1s, line 134: Looks a better name, wdyt ?

```suggestion
        _injected_files: Vec<FileMap>,
```

crates/provider/src/native.rs
```diff
@@ -153,9 +144,9 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
 
     async fn spawn_temp(
         &self,
-        node: Node,
-        files_inject: Vec<FileMap>,
-        files_get: Vec<FileMap>,
+        _node: Node,
+        _files_inject: Vec<FileMap>,
```
l0r1s, line 148:
```suggestion
        _injected_files: Vec<FileMap>,
```

crates/provider/src/native.rs
```diff
@@ -153,9 +144,9 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+        _files_get: Vec<FileMap>,
```
l0r1s, line 149: What is the goal of this input ?

crates/provider/src/native.rs
```diff
@@ -296,19 +287,91 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            ))?;
+
+        let _ = self
+            .run_command(
+                vec![format!("kill -STOP {}", process.pid)],
+                NativeRunCommandOptions {
+                    is_failure_allowed: true,
+                },
+            )
+            .await?;
```
l0r1s, line 307: Do we have some way to log ? It could be a good idea to debug log the Result here

crates/provider/src/native.rs
```diff
@@ -296,19 +287,91 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            ))?;
+
+        let _ = self
+            .run_command(
+                vec![format!("kill -CONT {}", process.pid)],
+                NativeRunCommandOptions {
+                    is_failure_allowed: true,
+                },
+            )
+            .await?;
```
l0r1s, line 328: Same here, if we have some definied way to log

crates/provider/src/native.rs
```diff
@@ -296,19 +287,91 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
-        // use &mut here since we mostly need to update the the pid/maps
+    // TODO: Add test
+    async fn restart(&mut self, node_name: &str, _after_secs: u16) -> Result<bool, ProviderError> {
+        let process = self
+            .process_map
+            .get(node_name)
+            .ok_or(ProviderError::MissingNodeInfo(
+                node_name.to_owned(),
+                "process".into(),
+            ))?;
```
l0r1s, line 340: I see this code defined multiple time, maybe we could extract a method ?

```suggestion
fn get_node_from_name(&self, name: &str) -> Result<bool, ProviderError> {
        self
            .process_map
            .get(node_name)
            .ok_or(ProviderError::MissingNodeInfo(
                node_name.to_owned(),
                "process".into(),
            ))
}
```

We could have the same for ```get_node_from_name_mut```

crates/provider/src/native.rs
```diff
@@ -296,19 +287,91 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            .map_err(|e| ProviderError::ErrorSpawningNode(e.to_string()))?;
+
+        let pid = if let Some(pid) = child_process.id() {
+            pid
+        } else {
+            return Err(ProviderError::ErrorSpawningNode(
+                "Failed to get pid".to_string(),
+            ));
+        };
+        process.pid = pid;
```
l0r1s, line 373: process.pid = ch
```suggestion
        process.pid = child_process.id()
                .ok_or(
                    Err(ProviderError::SpawningNode("Failed to get pid".to_string())
                 )?;
```

l0r1s (commented):
Small comments to have less duplication, else lgtm :+1:

crates/provider/src/native.rs
wirednkod: Doesn't really matter for this PR. Once implementation is done we can address the name based on usability

crates/provider/src/native.rs
wirednkod: same as above

crates/provider/src/native.rs
wirednkod: Doesn't really matter for this PR. Once implementation is done we can address the name based on usability

crates/provider/src/native.rs
wirednkod: we do not at this point afaik

crates/provider/src/native.rs
wirednkod: we do not at this point afaik

crates/provider/src/native.rs
pepoviola: I have originally `files_to_inject` / `files_to_get` but I found very verbose. Now I think those are better names to explain what is the reason behind those `inputs`.

crates/provider/src/native.rs
pepoviola: I will add the `log` crate, since we want to use later the `env_logger`

crates/provider/src/native.rs
pepoviola: Yes, make sense to extract that method to use it internally.

crates/provider/src/native.rs
```diff
@@ -296,19 +287,86 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+        let process: &mut Process =
+            self.process_map
+                .get_mut(node_name)
+                .ok_or(ProviderError::MissingNodeInfo(
+                    node_name.to_owned(),
+                    "process".into(),
+                ))?;
+
+        let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
+
```
pepoviola, line 358: We should `sleep` for `after_secs` here before restart the process.

crates/provider/src/native.rs
```diff
@@ -296,19 +287,86 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+            )
+            .await?;
         Ok(())
     }
 
-    async fn restart(&mut self, node_name: &str, after_secs: u16) -> Result<bool, ProviderError> {
-        // TODO: impl using run_command
-        // use &mut here since we mostly need to update the the pid/maps
+    // TODO: Add test
+    async fn restart(&mut self, node_name: &str, _after_secs: u16) -> Result<bool, ProviderError> {
```
pepoviola, line 333:
```suggestion
    async fn restart(&mut self, node_name: &str, after_secs: Option<u16>) -> Result<bool, ProviderError> {
```

crates/provider/src/native.rs
```diff
@@ -296,19 +299,76 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+        self.run_command(
+            vec![format!("kill -9 {:?}", process.pid)],
+            NativeRunCommandOptions {
+                is_failure_allowed: true,
+            },
+        )
+        .await?;
+
+        let sec: u64 = after_secs.unwrap_or(5).into();
+        sleep(Duration::from_millis(sec * 1000)).await;
```
pepoviola, line 349: if let Some(secs) = after_secs {
    sleep(Duration::from_secs(secs)).await;
}

crates/provider/src/native.rs
```diff
@@ -80,6 +74,15 @@ impl<T: FileSystem + Send + Sync> NativeProvider<T> {
             filesystem,
         }
     }
+
+    pub fn get_node_from_name(&self, node_name: &str) -> Result<&Process, ProviderError> {
```
pepoviola, line 78:
```suggestion
   fn get_process_by_node_name(&self, node_name: &str) -> Result<&Process, ProviderError> {
```
I think this is a better naming, since we are getting a ref of the process (by node name). Also, this method doesn't need to be public since will only be used internally.

crates/provider/src/native.rs
```diff
@@ -296,19 +299,77 @@ impl<T: FileSystem + Send + Sync> Provider for NativeProvider<T> {
...
+
+        let process: &mut Process =
+            self.process_map
+                .get_mut(node_name)
+                .ok_or(ProviderError::MissingNodeInfo(
+                    node_name.to_owned(),
+                    "process".into(),
+                ))?;
+
+        let mapped_env: HashMap<String, String> = env::vars().map(|(k, v)| (k, v)).collect();
```
pepoviola, line 360: The idea here is to add `our` env vars to the process, we should get those from the original process (if were set).

pepoviola: Actually, this will require a little more of work since we are not even storing the original `ProcessEnvironment` in the map, let commented out for now and I will add in a follow-up pr.

pepoviola (approved):

PR #84, opened 2023-06-26
add ChainDefaultContext to pass the default context to node config bu…

pepoviola:
…ilder

---
- Add struct `ChainDefaultContext` to hold the chain context defaults to be passed to the node config builder.
- Add method in `relaychain` config builder to get the `default_context` (needs docs)
- Change in `tests` to use `ChainDefaultContext::Default` (derived).

### DIFF
--- a/crates/configuration/src/shared/node.rs
+++ b/crates/configuration/src/shared/node.rs
@@ -7,7 +7,7 @@ use super::{
     helpers::{merge_errors, merge_errors_vecs},
     macros::states,
     resources::ResourcesBuilder,
-    types::{AssetLocation, Command, Image},
+    types::{AssetLocation, ChainDefaultContext, Command, Image},
 };
 use crate::shared::{
     resources::Resources,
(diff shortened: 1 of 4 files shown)


wirednkod (approved):
code looks much cleaner now 🚀

crates/configuration/src/shared/node.rs
```diff
@@ -206,20 +206,14 @@ impl<A> NodeConfigBuilder<A> {
...
 
 impl NodeConfigBuilder<Initial> {
-    pub fn new(
-        default_command: Option<Command>,
-        default_image: Option<Image>,
-        default_resources: Option<Resources>,
-        default_db_snapshot: Option<AssetLocation>,
-        default_args: Vec<Arg>,
-    ) -> Self {
+    pub fn new(chain_default_context: ChainDefaultContext) -> Self {
```
l0r1s, line 209: I think it makes it clearer, no ?

```suggestion
    pub fn new(context: ChainDefaultContext) -> Self {
```

crates/configuration/src/shared/node.rs
```diff
@@ -206,20 +206,14 @@ impl<A> NodeConfigBuilder<A> {
...
-                command: default_command,
-                image: default_image,
-                resources: default_resources,
-                db_snapshot: default_db_snapshot,
-                args: default_args,
+                command: chain_default_context.default_command,
+                image: chain_default_context.default_image,
+                resources: chain_default_context.default_resources,
+                db_snapshot: chain_default_context.default_db_snapshot,
+                args: chain_default_context.default_args,
```
l0r1s, line 216: With the upper one

```suggestion
                command: context.default_command,
                image: context.default_image,
                resources: context.default_resources,
                db_snapshot: context.default_db_snapshot,
                args: context.default_args,
```

pepoviola:
Updated, with your suggestions @l0r1s. Thanks!

PR #85, opened 2023-06-27
Feat/parachain defaults

l0r1s:
Add defaults methods to ParachainConfigBuilder for db_snapshot, image, command, args and resources, refacto tests and add constraint so Chain can't be empty

Coverage:
Screenshot from 2023-06-27 19-55-45


crates/configuration/src/parachain.rs
```diff
@@ -23,14 +24,24 @@ pub struct ParachainConfig {
...
+    default_command: Option<Command>,
+
+    default_image: Option<Image>,
+
+    default_resources: Option<Resources>,
+
+    default_db_snapshot: Option<AssetLocation>,
+
+    default_args: Vec<Arg>,
+
```
l0r1s, line 44: I didn't put comment here voluntary, the refacto of the doc will come in the next PR fyi

crates/configuration/src/parachain.rs
```diff
@@ -23,14 +24,24 @@ pub struct ParachainConfig {
     id: u32,
 
     /// Chain to use (use None if you are running adder-collator or undying-collator).
-    chain: Option<Chain>,
+    chain: Chain,
```
pepoviola, line 27: If we will use an empty string as no-chain, used for example with adder/undying collators, we should document the behavior.
Thx!

crates/configuration/src/network.rs
```diff
@@ -393,17 +397,20 @@ mod tests {
             .with_parachain(|parachain1| {
                 parachain1
                     .with_id(1000)
+                    .with_chain("myparachain1")
```
wirednkod, line 400:
```suggestion
            .with_parachain(|parachain| {
                parachain
                    .with_id(1000)
                    .with_chain("myparachain1")
```

crates/configuration/src/network.rs
```diff
@@ -393,17 +397,20 @@ mod tests {
...
                     .with_collator(|collator| {
                         collator
                             .with_name("collator")
                             .with_command("invalid command")
                             .validator(true)
                     })
-                    .with_initial_balance(100_000)
             })
             .with_parachain(|parachain2| {
                 parachain2
```
wirednkod, line 410: closure is only inside the `( )` scope so no need to alter the name every time
```suggestion
            .with_parachain(|parachain| {
                parachain
```

crates/configuration/src/parachain.rs
```diff
@@ -124,9 +136,16 @@ impl Default for ParachainConfigBuilder<Initial> {
         Self {
             config: ParachainConfig {
                 id: 100,
-                chain: None,
+                chain: ""
+                    .try_into()
+                    .expect("empty string should be valid. this is a bug, please report it: https://github.com/paritytech/zombienet-sdk/issues"),
```
wirednkod, line 141:
```suggestion
                    .expect("Empty string should be valid. This is a bug, please report it at: https://github.com/paritytech/zombienet-sdk/issues"),
```

crates/configuration/src/parachain.rs
```diff
@@ -412,6 +513,18 @@ mod tests {
...
+                    .with_name("collator1")
+                    .with_command("command1")
+                    .bootnode(true)
+            })
+            .with_collator(|collator2| {
+                collator2
+                    .with_name("collator2")
+                    .with_command("command2")
+                    .validator(true)
+            })
```
wirednkod, line 527:
```suggestion
            .with_collator(|collator| {
                collator
                    .with_name("collator1")
                    .with_command("command1")
                    .bootnode(true)
            })
            .with_collator(|collator| {
                collator
                    .with_name("collator2")
                    .with_command("command2")
                    .validator(true)
            })
```

crates/configuration/src/parachain.rs
pepoviola: I think we should discuss about this, since if we allow to set an `empty` string as chain value we should handle that internally since will fail at running command like 
```
polkadot build-spec --chain
```

What are the `pros` on allow empty strings as valid `chain`?

l0r1s: The config can't be built with an empty chain, this is only to fullfill the initial values but the typestate prevent to have non defined chain, or at least it should be added at validation, atm we only have no spaces validation

pepoviola: Ahh oks, thanks for the clarification. Also, I think we should add the check into validation step 👍 .

Thanks! (I will mark as resolved)

crates/configuration/src/network.rs
l0r1s: Yes but I think it's better for readability + only present in tests so that's fine imho

wirednkod: I think its unecessery. Preserving the same pattern makes readability easier. not a blocking for the approval - just a proposal

l0r1s: Done

pepoviola (approved):

pepoviola:
Looks good, thanks @l0r1s. I think we can automate the coverage report for each `pr` as a separate pr.

Thx!

l0r1s:
Yes, I was thinking about it but we need to store the coverage value somewhere to be able to compare between each PR

pepoviola:
> Yes, I was thinking about it but we need to store the coverage value somewhere to be able to compare between each PR

Great, I will merge now and add a new issue to implement the coverage in each pr.
Thanks!

PR #92, opened 2023-08-02
Feat/missing onboard as parachain field

l0r1s:
The field ```onboard_as_parachain``` is missing on the ParachainConfig, which enables the user to tells if he wants the registered parathread (using genesis or extrinsic at spawn time) to be promoted to a parachain.

By default this value is true, hence skipped in the dump.

### DIFF
--- a/crates/configuration/src/parachain.rs
+++ b/crates/configuration/src/parachain.rs
@@ -43,6 +43,7 @@ pub struct ParachainConfig {
     chain: Option<Chain>,
     #[serde(flatten)]
     registration_strategy: Option<RegistrationStrategy>,
+    #[serde(skip_serializing_if = "super::utils::is_true")]
     onboard_as_parachain: bool,
     #[serde(rename = "balance")]
     initial_balance: U128,
@@ -998,5 +999,17 @@ mod tests {
             errors.get(4).unwrap().to_string(),
             "parachain[2000].collators['collator2'].image: 'invalid.image' doesn't match regex '^([ip]|[hostname]/)?[tag_name]:[tag_version]?$'"
         );
+    }
+
+    #[test]
+    fn onboard_as_parachain_should_default_to_true() {
+        let config = ParachainConfigBuilder::new(Default::default())
+            .with_id(2000)
+            .with_chain("myparachain")
+            .with_collator(|collator| collator.with_name("collator"))
+            .build()
+            .unwrap();
+
+        assert!(config.onboard_as_parachain());
     }
 }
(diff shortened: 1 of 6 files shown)


crates/configuration/testing/snapshots/0001-big-network.toml
```diff
@@ -63,6 +64,7 @@ balance = 1000000000
 id = 2000
 chain = "myotherparachain"
 add_to_genesis = true
+onboard_as_parachain = true
```
wirednkod, line 67: since default is `true` this should not be needed

crates/configuration/testing/snapshots/0002-overridden-defaults.toml
```diff
@@ -50,6 +50,7 @@ cpu = "5Gi"
 id = 1000
 chain = "myparachain"
 add_to_genesis = true
+onboard_as_parachain = true
```
wirednkod, line 53: same as above

crates/configuration/src/parachain.rs
```diff
@@ -656,6 +676,7 @@ mod tests {
             parachain_config.registration_strategy().unwrap(),
             &RegistrationStrategy::UsingExtrinsic
         );
+        assert!(!parachain_config.onboard_as_parachain());
```
wirednkod, line 679: We need one more test where the `onboard_as_parachain` will not be set and the assertion will ensure that the `bool` value is `true` by default

crates/configuration/testing/snapshots/0001-big-network.toml
pepoviola: > since default is `true` this should not be needed

That true, but I think tracking if the user `set`/`not set` the value to add or not the field to the `toml` is add more complexity, and since this fields will be calculated later in the `ComputeChainSpec` having this value in the `toml` is ok. I mean, if we `skip` the field if the value is `true` we will need to add an extra state tracking to check if the `true` is the default value or the value set by the user, since if the the user `set` the value we should show it in the config. TLDR, I think is ok leave the field.

l0r1s: @wirednkod @pepoviola thanks for the feedback, that's true that we could skip it using a custom function and the following serde macro attribute like this: ```#[skip_if(is_true)]```

wdyt ?

wirednkod: I think it is valid - allowing less typing and config where not needed. since it will be clearly mentioned in documentation, I think this could work

pepoviola: @l0r1s / @wirednkod, oks we can skip it for now and check the feedback in the future. Also, just to add some context, this config `dumping` is a helper that we want to use as integration and is not the main path for endusers of the sdk.
Thanks!

pepoviola (approved):

PR #95, opened 2023-08-08
Create and publish SDK docs in gh-pages

### LINKED ISSUE #94: Rust Docs
Research around the rust docs, and fix the configuration, in order to be able to generate docs without the dependencies (and maybe publish them to github-pages as well?).

wirednkod:
Fixes #94


pepoviola (approved):

l0r1s:
Hey, nice job, do we really need to commit all assets including build files ?

pepoviola:
> Hey, nice job, do we really need to commit all assets including build files ?

That a good question, should be a way to publish those without having them committed (at least not in main branch).

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,35 @@
+name: Cargo Create Docs
+
+on:
+  push:
+    branches: [nik-add-docs-in-CI]
```
l0r1s, line 5: Just modify this to "main" and should be good for merge :+1:

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request:
+    branches: [nik-add-docs-in-CI]
```
l0r1s, line 7: Same here

l0r1s (approved):
lgtme ! just changing the branches and ready to merge :)

.github/workflows/documentation.yml
wirednkod: still testing.. this is not working

.github/workflows/documentation.yml
wirednkod: still testing.. this is not working

PR #96, opened 2023-08-08
Refactor of ```Provider``` trait into ```Provider```, ```ProviderNamespace``` and ```ProviderNode```

pepoviola:
WIP refactor provider trait (scoped changes)


l0r1s:
## Support crate
* Refactored MockFileSystem into InMemoryFilesystem, simplified and tested.
* Added LocalFileSystem implementation tested.

## Provider crate
* Refactored Provider trait and split into 3 traits, Provider, ProviderNamespace and ProviderNode
* Added implementation of NativeProvider, NativeProviderNamespace and NativeProviderNode:
   * Use of the Inner pattern to hide RwLock behind a cheaply clonable struct and make it thread safe to operate on individual object
   * Implemented logic to poll logs of stdout/stderr using additional tasks and communication over a channel, making it easier to be tested by using the InMemoryFilesystem
   * Implemented logic to spawn, pause, resume & restart a node without relying on ```ps``` command but by using the ```nix``` traits (an equivalent exists for windows if needed to do conditional compilation of these features in the future)

.github/workflows/ci.yml
```diff
@@ -38,7 +38,8 @@ jobs:
         run: cargo build
 
       - name: Tests
-        run: cargo test
+        # there should be a unique test thread for native provider tests (asserting spawned processes count)
+        run: cargo test -- --test-threads 1
```
pepoviola, line 42: Maybe we can scope this to not run all our test in a single threaded env.

.github/workflows/ci.yml
```diff
@@ -60,7 +61,8 @@ jobs:
         uses: taiki-e/install-action@cargo-llvm-cov
 
       - name: Collect coverage data
-        run: cargo llvm-cov nextest --lcov --output-path lcov.info
+        # there should be a unique test thread for native provider tests (asserting spawned processes count)
+        run: cargo llvm-cov nextest -j 1 --lcov --output-path lcov.info
```
pepoviola, line 65: same here.

pepoviola (commented):
Looks great, just a couple of comments.

.github/workflows/ci.yml
wirednkod: What is the reason of threading all tests like this?

crates/provider/src/native.rs
```diff
@@ -1,833 +1,1522 @@
...
+        let base_dir_raw = format!("{}/{}", &self.base_dir.to_string_lossy(), &options.name);
+        let base_dir = PathBuf::from(&base_dir_raw);
+        let log_path = PathBuf::from(format!("{}/{}.log", base_dir_raw, &options.name));
+        let config_dir = PathBuf::from(format!("{}{}", base_dir_raw, NODE_CONFIG_DIR));
+        let data_dir = PathBuf::from(format!("{}{}", base_dir_raw, NODE_DATA_DIR));
+        let scripts_dir = PathBuf::from(format!("{}{}", base_dir_raw, NODE_SCRIPTS_DIR));
+        self.filesystem.create_dir(&base_dir).await?;
+        self.filesystem.create_dir(&config_dir).await?;
+        self.filesystem.create_dir(&data_dir).await?;
+        self.filesystem.create_dir(&scripts_dir).await?;
```
wirednkod, line 174: Do we need to await each one to happen? Can they all run in parallel and when all are done to fulfill the await? (something like `Promise.all` in JS)

crates/provider/src/native.rs
```diff
@@ -1,833 +1,1522 @@
...
+
+        // copy injected files
+        for file in options.injected_files {
```
wirednkod, line 177: same here.. we have 1 await for each loop that takes place.

crates/provider/src/native.rs
```diff
@@ -1,833 +1,1522 @@
...
-        script_path: String,
-        args: Vec<String>,
-    ) -> Result<RunCommandResponse, ProviderError> {
-        let script_filename = Path::new(&script_path)
-            .file_name()
-            .ok_or(ProviderError::InvalidScriptPath(script_path.clone()))?
-            .to_str()
-            .ok_or(ProviderError::InvalidScriptPath(script_path.clone()))?;
-        let script_path_in_pod = format!("{}/{}/{}", self.tmp_dir, identifier, script_filename);
+    async fn static_setup(&self) -> Result<(), ProviderError> {
```
wirednkod, line 268: Add a TODO Note maybe on what needs to be done?

wirednkod (commented):
Some minor questions only

.github/workflows/ci.yml
l0r1s: The reason is because we need to assert the "dummy_process" is running at some point and in some state (paused/restarted/resumed), and if there are multiple tests asserting this (restart, spawn_node, etc...), it will make the test fails, that is what the command refers to.

The performance loss is not that much given the 89 tests of configuration crate run in 0.02s. Only the provider tests takes a bit longer because we assert logs are being/not being written at some point for some time (5/10s) or a process is not present/has restarted (5/10s), the provider tests takes around 1 min.

It takes around 1m30 to run all tests on my machine.

I have a todo to optimize provider tests to run same process with different name.

crates/provider/src/native.rs
l0r1s: Good catch ! Here we can use [join!](https://rust-lang.github.io/async-book/06_multiple_futures/02_join.html) but only for the 3 lasts call because the base_dir needs to be create all 3.

crates/provider/src/native.rs
l0r1s: Fixed

.github/workflows/ci.yml
l0r1s: Fixed

crates/provider/src/native.rs
pepoviola: Great! we can also in the future use `map` for this cases to produce the vector of futures.
👍

crates/provider/src/native.rs
l0r1s: I added a small comment :+1:

wirednkod (approved):

PR #99, opened 2023-09-03
Refactor/provider trait l0r1s

l0r1s:


crates/provider/src/lib.rs
```diff
@@ -2,85 +2,126 @@ mod errors;
...
+        remote_src: PathBuf,
+        local_dest: PathBuf,
+    ) -> Result<(), ProviderError>;
 
-    // TODO(team): Do we need at this point to handle cleanner/pod-monitor?
+    async fn pause(&self) -> Result<(), ProviderError>;
+
+    async fn resume(&self) -> Result<(), ProviderError>;
+
+    async fn restart(&mut self, after: Option<Duration>) -> Result<(), ProviderError>;
```
pepoviola, line 122: In a second thought, the only reason to this be a `mut ref` is because we want to store the `pid` of the process in the native provider. Maybe we can store elsewhere in the native provider and don't restrict the rest of them.

crates/provider/src/native.rs
```diff
@@ -2,531 +2,920 @@ use std::{
...
+    filesystem: FS,
+    tmp_dir: Option<String>,
+}
 
-    fn require_image() -> bool {
-        false
-    }
+#[derive(Debug)]
+struct NativeProviderInner<FS: FileSystem + Send + Sync + Clone> {
+    capabilities: ProviderCapabilities,
```
pepoviola, line 48: Do we need to add this fields into the `inner`, I think `capabilities`, `tmp_dir` and `filesystem` could be part of the `NativeProvider` and only put the namespace into the inner struct, since those are mostly initialized in `new` and will not change.

crates/provider/src/native.rs
```diff
@@ -2,531 +2,920 @@ use std::{
...
-
-            let _ = self
-                .run_command(
-                    [format!("kill -9 {}", pids_to_kill.join(" "))].to_vec(),
-                    NativeRunCommandOptions {
-                        is_failure_allowed: true,
-                    },
-                )
-                .await?;
+                tmp_dir: options.tmp_dir.unwrap_or(DEFAULT_TMP_DIR.to_string()),
```
pepoviola, line 71: We should get the `tmp_dir` from the OS if is not provided.

crates/provider/src/native.rs
```diff
@@ -2,531 +2,920 @@ use std::{
...
+#[async_trait]
+impl<FS: FileSystem + Send + Sync + Clone + 'static> ProviderNode for NativeNode<FS> {
+    async fn name(&self) -> String {
+        self.inner.read().await.name.clone()
+    }
 
-        process.pid = child_process.id().ok_or(ProviderError::ErrorSpawningNode(
-            "Failed to get pid".to_string(),
-        ))?;
+    async fn endpoint(&self) -> Result<(IpAddr, Port), ProviderError> {
```
pepoviola, line 244: Ahh this will require holding a map of the ports, at least for `rpc`/`prometheus`. If we don't want to bloat the `NodeProvider` we should store the info at `ns` level.

crates/provider/src/native.rs
```diff
@@ -2,531 +2,920 @@ use std::{
...
+        Ok(())
     }
 
-    #[tokio::test]
-    async fn test_fielsystem_usage() {
-        let mut native_provider: NativeProvider<MockFilesystem> =
-            NativeProvider::new("something", "/tmp", MockFilesystem::new());
+    async fn resume(&self) -> Result<(), ProviderError> {
+        let inner = self.inner.write().await;
+        let raw_pid = inner.process.id().unwrap();
```
pepoviola, line 302: we should not `unwrap` here.

crates/provider/src/lib.rs
l0r1s: Yes, it's a mistake, doesn't need to be mutable because we use an Arc here in ```inner``` and data inside it can be borrowed as mutable (we store the lock)

crates/provider/src/native.rs
l0r1s: What do you mean, is there a way to detect the "default tmp dir" for an os ?

crates/provider/src/native.rs
l0r1s: Good catch, still some work to do on error handling !

PR #100, opened 2023-09-05
Get a config from toml and create a network

### LINKED ISSUE #97: Configuration toml to network
Reverse the dump_to_toml function in order tto create form a toml file networks;

wirednkod:
Fixes #97

### DIFF
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -12,6 +12,7 @@ use crate::{
         constants::{NO_ERR_DEF_BUILDER, RELAY_NOT_NONE, RW_FAILED, THIS_IS_A_BUG, VALID_REGEX},
         helpers::merge_errors_vecs,
         macros::states,
+        node::NodeConfig,
         types::{Arg, ValidationContext},
     },
 };
@@ -111,8 +112,15 @@ impl NetworkConfig {
             .cloned()
             .collect();
 
+        let mut nodes: Vec<NodeConfig> = network_config
+            .relaychain()
+            .nodes()
+            .into_iter()
+            .cloned()
+            .collect();
+
         // SAFETY: is ok to use `unwrap` here since we ensure that is some at the begging of this fn
-        for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+        for node in nodes.iter_mut() {
             if relaychain_default_command.is_some() {
                 // we modify only nodes which don't already have a command
                 if node.command.is_none() {
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -82,10 +82,6 @@ impl NetworkConfig {
             Err(anyhow!("Relay chain does not exist."))?
         }
 
-        // println!("+++++++++++++++++++++++++++");
-        // println!("{:?}", network_config);
-        // println!("+++++++++++++++++++++++++++");
-
         // retrieve the defaults relaychain for assigning to nodes if needed
         let relaychain_default_command: Option<Command> = network_config
             .relaychain
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -55,6 +55,109 @@ impl NetworkConfig {
 
         Ok(re.replace_all(&toml_string, "$u128_value").to_string())
     }
+
+    pub fn load_from_toml(path: &str) -> Result<NetworkConfig, toml::ser::Error> {
+        let file_str = fs::read_to_string(path).unwrap();
+        let re: Regex = Regex::new(r"(?<field_name>(initial_)?balance)\s+=\s+(?<u128_value>\d+)")
+            .expect(&format!("{} {}", VALID_REGEX, THIS_IS_A_BUG));
+        let mut network_config: NetworkConfig = toml::from_str(
+            re.replace_all(&file_str, "$field_name = \"$u128_value\"")
+                .as_ref(),
+        )
+        .unwrap();
+
+        if network_config.relaychain.is_none() {
+            // TODO: (nikos) handle case where relaychain is None which is not valid
+        }
+
+        // retrieve the defaults relaychain for assigning to nodes if needed
+        let relaychain_default_command = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_command()
+            .cloned();
+
+        let relaychain_default_image = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_image()
+            .cloned();
+
+        let relaychain_default_resources = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_resources()
+            .cloned();
+
+        let relaychain_default_db_snapshot = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_db_snapshot()
+            .cloned();
+
+        let default_args = network_config.relaychain().default_args.clone();
+
+        // if there is some existing defaults and
+        if relaychain_default_command.is_some() {
+            // here we've got our own clone of Option<Command> so we just unwrap it because we asserted above it exists
+            let default_command: Option<crate::shared::types::Command> = relaychain_default_command;
+            // we take all nodes as mutables
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                // we modify only nodes which don't arleady have a command
+                if node.command.is_none() {
+                    node.command = default_command.clone();
+                    node.chain_context.default_command = default_command.clone();
+                }
+            }
+        }
+
+        if relaychain_default_image.is_some() {
+            let default_image = relaychain_default_image;
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                if node.image.is_none() {
+                    node.image = default_image.clone();
+                    node.chain_context.default_image = default_image.clone();
+                }
+            }
+        }
+
+        if relaychain_default_resources.is_some() {
+            let default_resources = relaychain_default_resources;
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                if node.resources.is_none() {
+                    node.resources = default_resources.clone();
+                    node.chain_context.default_resources = default_resources.clone();
+                }
+            }
+        }
+
+        if relaychain_default_db_snapshot.is_some() {
+            let default_db_snapshot = relaychain_default_db_snapshot;
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                if node.db_snapshot.is_none() {
+                    node.db_snapshot = default_db_snapshot.clone();
+                    node.chain_context.default_db_snapshot = default_db_snapshot.clone();
+                }
+            }
+        }
+
+        if !default_args.is_empty() {
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                if node.args.is_empty() {
+                    node.args.clone_from(&default_args.clone());
+                    node.chain_context
+                        .default_args
+                        .clone_from(&default_args.clone());
+                }
+            }
+        }
+
+        Ok(network_config)
+    }
 }
 
 states! {
@@ -178,97 +281,6 @@ impl<A> NetworkConfigBuilder<A> {
 impl NetworkConfigBuilder<Initial> {
     pub fn new() -> NetworkConfigBuilder<Initial> {
         Self::default()
-    }
-
-    pub fn toml_to_network(path: &str) -> Result<NetworkConfig, toml::ser::Error> {
-        let file_str = fs::read_to_string(path).unwrap();
-
-        let mut network_config: NetworkConfig = toml::from_str(&file_str).unwrap();
-
-        if network_config.relaychain.is_none() {
-            // TODO: (nikos) handle case where relaychain is None which is not valid
-        }
-
-        // retrieve the defaults relaychain for assigning to nodes if needed
-        let relaychain_default_command = network_config
-            .relaychain
-            .as_ref()
-            .unwrap()
-            .default_command()
-            .cloned();
-
-        let relaychain_default_image = network_config
-            .relaychain
-            .as_ref()
-            .unwrap()
-            .default_image()
-            .cloned();
-
-        let relaychain_default_resources = network_config
-            .relaychain
-            .as_ref()
-            .unwrap()
-            .default_resources()
-            .cloned();
-
-        let relaychain_default_db_snapshot = network_config
-            .relaychain
-            .as_ref()
-            .unwrap()
-            .default_db_snapshot()
-            .cloned();
-
-        let default_args = network_config.relaychain().default_args.clone();
-
-        // if there is some existing defaults and
-        if relaychain_default_command.is_some() {
-            // here we've got our own clone of Option<Command> so we just unwrap it because we asserted above it exists
-            let default_command: Option<crate::shared::types::Command> = relaychain_default_command;
-            // we take all nodes as mutables
-            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
-                // we modify only nodes which don't arleady have a command
-                if node.command.is_none() {
-                    node.command = default_command.clone();
-                }
-            }
-        }
-
-        if relaychain_default_image.is_some() {
-            let default_image = relaychain_default_image;
-            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
-                if node.image.is_none() {
-                    node.image = default_image.clone();
-                }
-            }
-        }
-
-        if relaychain_default_resources.is_some() {
-            let default_resources = relaychain_default_resources;
-            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
-                if node.resources.is_none() {
-                    node.resources = default_resources.clone();
-                }
-            }
-        }
-
-        if relaychain_default_db_snapshot.is_some() {
-            let default_db_snapshot = relaychain_default_db_snapshot;
-            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
-                if node.db_snapshot.is_none() {
-                    node.db_snapshot = default_db_snapshot.clone();
-                }
-            }
-        }
-
-        if !default_args.is_empty() {
-            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
-                if node.args.is_empty() {
-                    node.args.clone_from(&default_args.clone());
-                }
-            }
-        }
-
-        Ok(network_config)
     }
 
     /// Set the relay chain using a nested [`RelaychainConfigBuilder`].
@@ -948,8 +960,8 @@ mod tests {
 
     #[test]
     fn the_toml_config_should_be_imported_and_match_a_network() {
-        let toml_to_network =
-            NetworkConfigBuilder::toml_to_network("./testing/snapshots/0000-small-network.toml")
+        let load_from_toml =
+            NetworkConfigBuilder::load_from_toml("./testing/snapshots/0000-small-network.toml")
                 .unwrap();
 
         let expected = NetworkConfigBuilder::new()
@@ -978,7 +990,7 @@ mod tests {
             })
             .build()
             .unwrap();
-        assert_eq!(toml_to_network, expected);
+        assert_eq!(load_from_toml, expected);
     }
 
     #[test]
@@ -987,8 +999,7 @@ mod tests {
         // let converted_toml = toml::from_str::<toml::Value>(file_str.as_str()).unwrap();
 
         let config =
-            NetworkConfigBuilder::toml_to_network("./testing/snapshots/0000-small-network.toml")
-                .unwrap();
+            NetworkConfig::load_from_toml("./testing/snapshots/0000-small-network.toml").unwrap();
         println!("{:?}", config);
     }
 }
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -163,7 +163,16 @@ impl NetworkConfig {
             .set_nodes(nodes);
 
         // Validation checks for parachains
-        // TryInto::<Chain>::try_into(network_config.parachains().first().unwrap().chain().as_str())?;
+        network_config.parachains().iter().for_each(|parachain| {
+            let _ = TryInto::<Chain>::try_into(parachain.chain().unwrap().as_str());
+
+            if parachain.default_image().is_some() {
+                let _ = TryInto::<Image>::try_into(parachain.default_image().unwrap().as_str());
+            }
+            if parachain.default_command().is_some() {
+                let _ = TryInto::<Command>::try_into(parachain.default_command().unwrap().as_str());
+            }
+        });
 
         Ok(network_config)
     }
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -1,4 +1,4 @@
-use std::{any, cell::RefCell, fs, marker::PhantomData, rc::Rc};
+use std::{cell::RefCell, fs, marker::PhantomData, rc::Rc};
 
 use regex::Regex;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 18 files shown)


l0r1s (changes requested):

crates/configuration/src/network.rs
```diff
@@ -178,6 +180,97 @@ impl NetworkConfigBuilder<Initial> {
         Self::default()
     }
 
+    pub fn toml_to_network(path: &str) -> Result<NetworkConfig, toml::ser::Error> {
```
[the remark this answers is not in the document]
pepoviola: Agree, I think moving to `NetworkConfig` and call it `load_from_toml` would be nice. :)

pepoviola, line 183: We should also need some specialized Err (or we can use anyhow at the moment).

crates/configuration/src/network.rs
```diff
@@ -178,6 +180,97 @@ impl NetworkConfigBuilder<Initial> {
...
+        let file_str = fs::read_to_string(path).unwrap();
```
pepoviola, line 184: `unwrap` will panic

crates/configuration/src/relaychain.rs
```diff
@@ -12,20 +12,20 @@ use crate::shared::{
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct RelaychainConfig {
     chain: Chain,
     default_command: Option<Command>,
     default_image: Option<Image>,
     default_resources: Option<Resources>,
     default_db_snapshot: Option<AssetLocation>,
     #[serde(skip_serializing_if = "std::vec::Vec::is_empty")]
-    default_args: Vec<Arg>,
+    pub(crate) default_args: Vec<Arg>,
```
pepoviola, line 23: Do we need this to have access from the `crate`?

crates/configuration/src/relaychain.rs
```diff
@@ -12,20 +12,20 @@ use crate::shared::{
...
     chain_spec_path: Option<AssetLocation>,
     random_nominators_count: Option<u32>,
     max_nominations: Option<u8>,
     #[serde(skip_serializing_if = "std::vec::Vec::is_empty")]
-    nodes: Vec<NodeConfig>,
+    pub(crate) nodes: Vec<NodeConfig>,
```
pepoviola, line 28: Maybe we can replace this with a method like `set_nodes`

crates/configuration/src/network.rs
```diff
@@ -178,6 +180,97 @@ impl NetworkConfigBuilder<Initial> {
...
+            }
+        }
+
+        if !default_args.is_empty() {
+            for node in network_config.relaychain.as_mut().unwrap().nodes.iter_mut() {
+                if node.args.is_empty() {
+                    node.args.clone_from(&default_args.clone());
+                }
+            }
+        }
```
pepoviola, line 269: We are iterating over the `nodes` here multiple times, one for each option. I think we can rewrite in the other way and iterate one time and make the check inside the loop. (I can send you a pr)

crates/configuration/src/shared/types.rs
```diff
@@ -365,15 +365,51 @@ pub struct ValidationContext {
 
 #[derive(Default, Debug, Clone, PartialEq, Deserialize)]
 pub struct ChainDefaultContext {
-    pub(crate) default_command: Option<Command>,
```
pepoviola, line 287: Hi @wirednkod, I'm not fully understand the reason behind this change. This is an internal type that is not exposed through the config builder right?

crates/configuration/src/shared/constants.rs
```diff
@@ -0,0 +1,11 @@
+pub const VALID_REGEX: &str = "Regex should be valid";
+pub const BORROWABLE: &str = "Must be borrowable as mutable";
+pub const RELAY_NOT_NONE: &str = "Typestate should ensure the relaychain isn't None at this point,";
+pub const SHOULD_COMPILE: &str = "Should compile with success";
+pub const INFAILABLE: &str = "Infaillible";
+pub const NO_ERR_DEF_BUILDER: &str = "Should have no errors for default builder";
+pub const RW_FAILED: &str = "Should be able to read/write - failed";
+pub const OPT_UNWRAP_ERR: &str = "Option unwrap error";
```
pepoviola, line 8: Is this used? seems like not.

crates/configuration/src/relaychain.rs
pepoviola:
```suggestion
    nodes: Vec<NodeConfig>,
```
we can make it private again here since you added the `set_nodes` fn

crates/configuration/src/shared/types.rs
wirednkod: Correct. Since this is internal and as we agreed this we will keep as `pub(crate)`;

crates/configuration/src/parachain.rs
```diff
@@ -223,7 +229,7 @@ impl<A> ParachainConfigBuilder<A> {
             default_image: self.config.default_image.clone(),
             default_resources: self.config.default_resources.clone(),
             default_db_snapshot: self.config.default_db_snapshot.clone(),
-            default_args: self.config.default_args.clone(),
+            default_args: self.config.default_args().into_iter().cloned().collect(),
```
l0r1s, line 232: Not needed

```suggestion
            default_args: self.config.clone(),
```

crates/configuration/src/relaychain.rs
```diff
@@ -78,6 +79,11 @@ impl RelaychainConfig {
     pub fn nodes(&self) -> Vec<&NodeConfig> {
         self.nodes.iter().collect::<Vec<&NodeConfig>>()
     }
+
+    /// The nodes of the relay chain.
```
l0r1s, line 83: This code is not exposed publicly:

```suggestion
    // The nodes of the relay chain.
```

Given how straightforward is this code, we can even skip it ?

crates/configuration/src/relaychain.rs
```diff
@@ -135,11 +141,11 @@ impl<A> RelaychainConfigBuilder<A> {
 
     fn default_chain_context(&self) -> ChainDefaultContext {
         ChainDefaultContext {
-            default_command: self.config.default_command.clone(),
-            default_image: self.config.default_image.clone(),
+            default_command: self.config.default_command().cloned(),
+            default_image: self.config.default_image().cloned(),
```
l0r1s, line 145: Not needed, the getter are for the public API, we have access to the inner directly here:

```suggestion
            default_command: self.config.default_command.clone(),
            default_image: self.config.default_image.clone(),
```

crates/configuration/src/relaychain.rs
```diff
@@ -135,11 +141,11 @@ impl<A> RelaychainConfigBuilder<A> {
...
             default_resources: self.config.default_resources.clone(),
             default_db_snapshot: self.config.default_db_snapshot.clone(),
-            default_args: self.config.default_args.clone(),
+            default_args: self.config.default_args().into_iter().cloned().collect(),
```
l0r1s, line 148: Same

```suggestion
            default_args: self.config.default_args.clone(),
```

crates/configuration/src/shared/node.rs
```diff
@@ -152,16 +168,30 @@ impl NodeConfig {
...
+    pub fn set_image(&mut self, image: Image) {
+        self.image = Some(image);
+    }
+
     /// Command to run the node.
     pub fn command(&self) -> Option<&Command> {
         self.command.as_ref()
     }
 
+    pub fn set_command(&mut self, command: Command) {
```
l0r1s, line 181:
```suggestion
    pub(crate)  fn set_command(&mut self, command: Command) {
```

crates/configuration/src/shared/node.rs
```diff
@@ -152,16 +168,30 @@ impl NodeConfig {
         self.image.as_ref()
     }
 
+    /// The default container image used for nodes.
+    pub fn set_image(&mut self, image: Image) {
```
l0r1s, line 172: Be careful of what we expose as public API

```suggestion
    pub(crate) fn set_image(&mut self, image: Image) {
```

crates/configuration/src/shared/node.rs
```diff
@@ -95,7 +104,14 @@ impl Serialize for NodeConfig {
...
 
-        if self.args.is_empty() || self.args == self.chain_context.default_args {
+        if self.args.is_empty()
+            || self.args()
+                == self
+                    .chain_context
+                    .default_args
+                    .iter()
+                    .collect::<Vec<&Arg>>()
+        {
```
l0r1s, line 114: Why modify this code ? We don't need to use getter because we have access to self fields and PartialEq borrow automatically here: https://doc.rust-lang.org/nightly/std/cmp/trait.PartialEq.html

```suggestion
        if self.args.is_empty() || self.args == self.chain_context.default_args {
```

crates/configuration/src/shared/node.rs
```diff
@@ -52,27 +52,36 @@ impl From<(&str, &str)> for EnvVar {
...
 
 /// A node configuration, with fine-grained configuration options.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct NodeConfig {
     name: String,
-    image: Option<Image>,
-    command: Option<Command>,
+    pub(crate) image: Option<Image>,
+    pub(crate) command: Option<Command>,
```
l0r1s, line 59: If we use setter we don't need to expose the fields to the crate (set_image/set_command are defined below), we can go only setters or only public field to crate to make it uniform.

crates/configuration/src/relaychain.rs
```diff
@@ -101,7 +107,7 @@ impl Default for RelaychainConfigBuilder<Initial> {
             config: RelaychainConfig {
                 chain: "default"
                     .try_into()
-                    .expect("'default' overriding should be ensured by typestate. this is a bug, please report it: https://github.com/paritytech/zombienet-sdk/issues"),
+                    .unwrap_or_else(|_| panic!("{} {}", DEFAULT_TYPESTATE, THIS_IS_A_BUG)),
```
l0r1s, line 110: We should disable this lint from clippy and revert to using expect with &format: https://github.com/rust-lang/rust-clippy/issues/2928

crates/configuration/src/shared/helpers.rs
```diff
@@ -31,7 +32,7 @@ pub fn ensure_node_name_unique(
 ) -> Result<(), anyhow::Error> {
     let mut context = validation_context
         .try_borrow_mut()
-        .expect("must be borrowable as mutable, this is a bug please report it: https://github.com/paritytech/zombienet-sdk/issues");
+        .unwrap_or_else(|_| panic!("{}, {}", BORROWABLE, THIS_IS_A_BUG));
```
l0r1s, line 35: Same as above

crates/configuration/src/shared/helpers.rs
```diff
@@ -47,7 +48,7 @@ pub fn ensure_port_unique(
...
```
l0r1s, line 51: Same as above

crates/configuration/src/shared/resources.rs
```diff
@@ -43,7 +48,7 @@ impl TryFrom<&str> for ResourceQuantity {
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         lazy_static! {
             static ref RE: Regex = Regex::new(r"^\d+(.\d+)?(m|K|M|G|T|P|E|Ki|Mi|Gi|Ti|Pi|Ei)?$")
-                .expect("should compile with success. this is a bug, please report it: https://github.com/paritytech/zombienet-sdk/issues");
+                .unwrap_or_else(|_| panic!("{}, {}", SHOULD_COMPILE, THIS_IS_A_BUG));
```
l0r1s, line 51: Same as above

crates/configuration/src/shared/types.rs
```diff
@@ -27,6 +32,14 @@ impl From<u128> for U128 {
...
 }
 
+// impl TryFrom<&str> for U128 {
+//     type Error = ConversionError;
+
+//     fn try_from(value: &str) -> Result<Self, Self::Error> {
+//         Ok(Self(value.to_string().parse::<u128>().unwrap()))
+//     }
+// }
+
```
l0r1s, line 42: Why is is commented ? Could probably be reused below with some modification:

```suggestion
use std::error::Error;

impl TryFrom<&str> for U128 {
     type Error = Box<dyn Error>;

     fn try_from(value: &str) -> Result<Self, Self::Error> {
         Ok(Self(value.to_string().parse::<u128>()?))
     }
 }

```

crates/configuration/src/shared/types.rs
```diff
@@ -38,6 +51,32 @@ impl Serialize for U128 {
...
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("an integer between 0 and 2^128 − 1.")
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                Ok(U128(v.to_string().parse::<u128>().unwrap()))
```
l0r1s, line 72: With above modification we get:

```suggestion
                v.try_into().map_err(de::Error::custom)
```

crates/configuration/src/shared/types.rs
```diff
@@ -38,6 +51,32 @@ impl Serialize for U128 {
...
+            }
+        }
+
+        deserializer.deserialize_str(U128Visitor)
+    }
+}
```
l0r1s, line 78: I think it makes sense to put the visitor definition out of the Deserialize implementation:

```suggestion
struct U128Visitor;
        
impl<'de> de::Visitor<'de> for U128Visitor {
    type Value = U128;

    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        formatter.write_str("an integer between 0 and 2^128 − 1.")
    }

    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
    where
        E: de::Error,
    {
        Ok(U128(v.to_string().parse::<u128>().unwrap()))
    }
}

impl<'de> Deserialize<'de> for U128 {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        deserializer.deserialize_str(U128Visitor)
    }
}
```

crates/configuration/src/shared/types.rs
```diff
@@ -110,7 +149,7 @@ impl TryFrom<&str> for Image {
             static ref RE: Regex = Regex::new(&format!(
                 "^({IP_PART}|{HOSTNAME_PART}/)?{TAG_NAME_PART}(:{TAG_VERSION_PART})?$",
             ))
-            .expect("should compile with success. this is a bug, please report it: https://github.com/paritytech/zombienet-sdk/issues");
+            .unwrap_or_else(|_| panic!("{}, {}", SHOULD_COMPILE, THIS_IS_A_BUG));
```
l0r1s, line 152: Same as above

crates/configuration/src/shared/types.rs
```diff
@@ -208,7 +247,8 @@ impl From<&str> for AssetLocation {
         }
 
         Self::FilePath(
-            PathBuf::from_str(value).expect("infaillible. this is a bug, please report it"),
+            PathBuf::from_str(value)
+                .unwrap_or_else(|_| panic!("{}, {}", INFAILABLE, THIS_IS_A_BUG)),
```
l0r1s, line 251: Same as above

crates/configuration/src/shared/types.rs
```diff
@@ -231,6 +271,32 @@ impl Serialize for AssetLocation {
...
+            where
+                E: de::Error,
+            {
+                Ok(AssetLocation::from(v))
+            }
+        }
+
+        deserializer.deserialize_any(AssetLocationVisitor)
+    }
+}
```
l0r1s, line 298: Same as above, remove the AssetLocationVisitor + Visitor implem trait out of the Deserialize implem

crates/configuration/src/shared/types.rs
```diff
@@ -276,18 +342,59 @@ impl Serialize for Arg {
...
+                    }
+                } else {
+                    Ok(Arg::Flag(v.to_string()))
+                }
+            }
+        }
+
+        deserializer.deserialize_any(ArgVisitor)
+    }
+}
```
l0r1s, line 383: Same as above

crates/configuration/src/shared/types.rs
```diff
@@ -276,18 +342,59 @@ impl Serialize for Arg {
...
+                        let split: Vec<&str> = v.split(' ').collect::<Vec<&str>>();
+                        if split.len() == 1 {
+                            Ok(Arg::Flag(v.to_string()))
+                        } else {
+                            Ok(Arg::Option(split[0].to_string(), split[1].to_string()))
+                        }
+                    }
+                } else {
+                    Ok(Arg::Flag(v.to_string()))
+                }
```
l0r1s, line 377: I think this can be more readable using 3 cases (contains '=', starts_with '--'/'-', flag) so moving inner if else out is probably better for readability, another thing is that we are indexing the split array without knowing the size of the vector, it's unsafe and can panic, imagine the case where we do "--test=".split('=')[1], we panic.

crates/configuration/src/utils.rs
```diff
@@ -1,3 +1,7 @@
 pub fn is_true(value: &bool) -> bool {
     *value
 }
+
+pub fn default_as_true() -> bool {
+    true
+}
```
l0r1s, line 7: Asking myself if those functions can be made ```pub(crate)``` ?

l0r1s (changes requested):
Good job ! We probably should unit tests custom Serialize/Deserialize implems to be sure they work for every edge cases

crates/configuration/src/parachain.rs
wirednkod: That does not seem correct 👀 
Why clone the config in the default args (type mismatch)

crates/configuration/src/shared/node.rs
wirednkod: Will remove the setters. We will keep the pub(crate)

crates/configuration/src/parachain.rs
l0r1s: typo error ```self.config.default_args.clone()```

crates/configuration/src/utils.rs
wirednkod: yes they can 👍🏽

crates/configuration/src/shared/constants.rs
```diff
@@ -0,0 +1,15 @@
...
+pub const INFAILABLE: &str = "Infaillible. ";
+pub const NO_ERR_DEF_BUILDER: &str = "Should have no errors for default builder. ";
+pub const RW_FAILED: &str = "Should be able to read/write - failed. ";
+pub const DEFAULT_TYPESTATE: &str = "'default' overriding should be ensured by typestate.";
+pub const VALIDATION_CHECK: &str = "Validation failed. ";
+
+pub const PREFIX_CANT_BE_NONE: &str = "Name prefix can't be None if a value exists.";
+
+pub const THIS_IS_A_BUG: &str =
+    "This is a bug please report it: https://github.com/paritytech/zombienet-sdk/issues";
```
l0r1s, line 15: One thing we could agree on is having non-capitalized error message ? This is something common in Rust, wdyt ?

crates/configuration/src/shared/resources.rs
```diff
@@ -72,19 +77,19 @@ pub struct Resources {
     limit_cpu: Option<ResourceQuantity>,
 }
 
+#[derive(Serialize, Deserialize)]
+struct ResourcesField {
+    memory: Option<ResourceQuantity>,
+    cpu: Option<ResourceQuantity>,
+}
+
 impl Serialize for Resources {
```
l0r1s, line 86: :+1:

crates/configuration/src/shared/constants.rs
wirednkod: I am not aware of that. If its a "thing" we should do it

crates/configuration/src/network.rs
```diff
@@ -48,11 +58,117 @@ impl NetworkConfig {
...
+            .cloned();
+
+        let relaychain_default_image: Option<Image> = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_image()
+            .cloned();
+
+        let relaychain_default_db_snapshot: Option<crate::shared::types::AssetLocation> =
```
pepoviola, line 97:
```suggestion
        let relaychain_default_db_snapshot: Option<AssetLocation> =
```

this need to bring the type to the scope with 
```rs
use crate::shared::types::AssetLocation
```

crates/configuration/src/network.rs
```diff
@@ -48,11 +58,117 @@ impl NetworkConfig {
...
+            .relaychain
+            .as_ref()
+            .unwrap()
+            .default_command()
+            .cloned();
+
+        let relaychain_default_image: Option<Image> = network_config
+            .relaychain
+            .as_ref()
+            .unwrap()
```
pepoviola, line 93: We should annotate why is `safe` to use `unwrap` here (since we are checking `relaychain` is some before),
Also, I think this could be refactored in a nice `if let` :)

crates/configuration/src/network.rs
```diff
@@ -48,11 +58,117 @@ impl NetworkConfig {
...
+
+        let relaychain_default_db_snapshot: Option<crate::shared::types::AssetLocation> =
+            network_config
+                .relaychain
+                .as_ref()
+                .unwrap()
+                .default_db_snapshot()
+                .cloned();
+
+        let default_args: Vec<Arg> = network_config
```
pepoviola, line 105: We are using two ways here to access the data
`.relaychain.as_ref().unwrap()` and `.relaychain()` we should access in a uniform way here.

crates/configuration/src/network.rs
```diff
@@ -48,11 +58,117 @@ impl NetworkConfig {
...
+        if relaychain_default_image.is_some() {
+            TryInto::<Image>::try_into(relaychain_default_image.clone().expect(VALIDATION_CHECK))?;
+        }
+        if relaychain_default_command.is_some() {
+            TryInto::<Command>::try_into(
+                relaychain_default_command.clone().expect(VALIDATION_CHECK),
+            )?;
+        }
+
+        // SAFETY: is ok to use `unwrap` here since we ensure that is some at the begging of this fn
```
pepoviola, line 130:
```suggestion
        // SAFETY: is ok to use `unwrap` here since we ensure that is some at the begging of this fn
```
I think this comment is outdated (needs to be a few lines before :))

crates/configuration/src/network.rs
```diff
@@ -48,11 +58,117 @@ impl NetworkConfig {
...
+
+        network_config
+            .relaychain
+            .as_mut()
+            .expect(&format!("{}, {}", NO_ERR_DEF_BUILDER, THIS_IS_A_BUG))
+            .set_nodes(nodes);
+
+        // Validation checks for parachains
+        network_config.parachains().iter().for_each(|parachain| {
+            let _ = TryInto::<Chain>::try_into(parachain.chain().unwrap().as_str());
```
pepoviola, line 160: We can use `ok_or` and return a error here instead of `unwrap` and panic.

crates/configuration/src/shared/constants.rs
pepoviola: https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=message#error-types-are-meaningful-and-well-behaved-c-good-err

```
The error message given by the Display representation of an error type should be lowercase without trailing punctuation, and typically concise.
```
We can take this approach :)

pepoviola (commented):
Looks good @wirednkod!!! Couple of comments and nits

crates/configuration/src/network.rs
wirednkod: I convert all to `.relaychain()` as it makes more sense than using `unwrap()`

crates/configuration/src/network.rs
wirednkod: `if let` is not the best choice for this particular place you mention, as the variable is reused, and thus `cloned()` at least 2 times (and the rest ones), so I prefered to leave it as is.

pepoviola (approved):

PR #102, opened 2023-09-20
Handle unique validation when parachain id is same

wirednkod:
Checks if the parachain ID exists and if so error is sent;

Note: Should we increase the parachain ID in case it already exists instead of panicking?


pepoviola:
We should allow use the same `id` for multiple parachains. There is already a test in cumulus using this feature (https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/zombienet/tests/0005-migrate_solo_to_para.toml). I think we can use the same logic that we have in the current version where we use another `identifier` for the parachain we just append `-n` if is already used. Thoughts?

l0r1s:
> We should allow use the same `id` for multiple parachains. There is already a test in cumulus using this feature (https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/zombienet/tests/0005-migrate_solo_to_para.toml). I think we can use the same logic that we have in the current version where we use another `identifier` for the parachain we just append `-n` if is already used. Thoughts?

I'm good for adding the field but we should find a better name :roll_eyes:

pepoviola:
> > We should allow use the same `id` for multiple parachains. There is already a test in cumulus using this feature (https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/zombienet/tests/0005-migrate_solo_to_para.toml). I think we can use the same logic that we have in the current version where we use another `identifier` for the parachain we just append `-n` if is already used. Thoughts?
> 
> 
> 
> I'm good for adding the field but we should find a better name :roll_eyes: 

Yes, maybe something like uniq_id 🫣

PR #111, opened 2023-09-29
Feat/genesis overrides

l0r1s:
- Added genesis_overrides fields to RelaychainConfig and ParachainConfig builders and added logic to merge it into the specs being created


pepoviola (commented):
Looks good! we should check why tests are failing and adding an small test for the new fn would be good 🙌

pepoviola (approved):

PR #114, opened 2023-10-03
feat: add node methods for actions and metrics

pepoviola:
Add methods to NetworkNode:
- pause
- resume
- restart
- reports (get metrics from prometheus)
- assert (basic implementation, we need to add the cmp logic).

Todo: add test when we have the processManager mock


wirednkod (approved):
lgtm... just to make the build pass

PR #118, opened 2023-10-06
feat: add verifier to check readiness

pepoviola:
Add the verifier logic to ensure readiness of the nodes and remove all the `sleeps` in the current examples.
Some todos to make in follow up prs.

### DIFF
--- a/crates/orchestrator/src/network_helper/verifier.rs
+++ b/crates/orchestrator/src/network_helper/verifier.rs
@@ -1,0 +1,31 @@
+use std::time::Duration;
+
+use tokio::time::timeout;
+
+use crate::network::node::NetworkNode;
+
+pub async fn verify_nodes(nodes: &[&NetworkNode]) -> Result<(), anyhow::Error> {
+    timeout(Duration::from_secs(90), check_nodes(nodes))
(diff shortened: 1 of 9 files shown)


crates/orchestrator/src/network_helper/verifier.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::network::node::NetworkNode;
+
+pub async fn verify_nodes(nodes: &[&NetworkNode]) -> Result<(), anyhow::Error> {
+    timeout(Duration::from_secs(90), check_nodes(nodes))
+        .await
+        .map_err(|_| anyhow::anyhow!("one or more nodes are not ready!"))
+}
+
+// TODO: we should inject in someway the logic to make the request
+// in order to allow us to `mock` and easily test this.
```
l0r1s, line 14:
```suggestion
// TODO: we should inject in someway the logic to make the request
// in order to allow us to `mock` and easily test this.
// maybe moved to the provider with a NodeStatus, and some helpers like wait_running, wait_ready, etc... ? to be discussed
```

l0r1s (approved):

PR #119, opened 2023-10-06
Fix various bugs

wirednkod:
Based on the simple example, various bugs have been identified;
This PR resolves them

### DIFF
--- a/crates/configuration/src/shared/types.rs
+++ b/crates/configuration/src/shared/types.rs
@@ -357,7 +357,7 @@ impl<'de> de::Visitor<'de> for ArgVisitor {
     {
         // covers the "-lruntime=debug,parachain=trace" case
         // TODO: Make this more generic by adding the scenario in the regex below
-        if (v.starts_with("-l") || v.starts_with("-log")) && v.contains(",") {
+        if (v.starts_with("-l") || v.starts_with("-log")) && v.contains(',') {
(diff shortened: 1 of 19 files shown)


crates/configuration/src/shared/types.rs
```diff
@@ -355,9 +355,13 @@ impl<'de> de::Visitor<'de> for ArgVisitor {
     where
         E: de::Error,
     {
+        // covers the "-lruntime=debug,parachain=trace" case
+        // TODO: Make this more generic by adding the scenario in the regex below
+        if (v.starts_with("-l") || v.starts_with("-log")) && v.contains(',') {
+            return Ok(Arg::Flag(v.to_string()));
```
pepoviola, line 361: So basically we are set this as flag instead of traing to parse it right? Did you try if there is no "," like -lparachain=debug ??

wirednkod: You are right. This should be better parsed

pepoviola: Is ok to treated as flag, but I think maybe there is a bug if isn't contains a ","

wirednkod: I tested the case of not containing "," and indeed it breaks; As a result I removed that `||` and retested the possible scenarions and now in both cases a `Flag` (as needed) returns

pepoviola (approved):

PR #121, opened 2023-10-08
Convert the dynamic builder to the more type-secure static one

wirednkod:
While running a local network I retrieved the metadata for the commands we care for using the command:
```sh
subxt metadata --url ws://127.0.0.1:44357 --pallets ParasSudoWrapperCall,Sudo,ValidatorManager > metadata.scale
```
I could have download all the metadata, but this saves some space;


wirednkod:
we decided that it makes more sense to keep the dynamic version for registering parachains and validators. Thus this will close

PR #122, opened 2023-10-09
Feat/process manager

l0r1s:
- Added an abstraction over the ```ProcessManager``` (```Command::spawn```, ```Command::output```) to make it easier to unit test the native provider
- Added a ```FakeProcessManager``` implementation with deterministic process advancing and log emitting through the ```advance_by(cycles)``` method, added a way to use dynamic logging based on passed environment, program and args for Stdout and Stderr
- Refactored all tests of the ```NativeProvider``` to make them platform independant
- Added ```OsProcessManager``` wrapping the tokio::process
- Refactored the ```FileSystem``` trait and implementations to use ```<T> and where T: MyTrait``` instead of ```impl MyTrait``` in parameters to uniformize across the crate.

Integration tests with real process manager will be moved to a new crate and they will be run only in pipeline to avoid issues and differences between platform (Macos/Unix)


crates/orchestrator/src/spawner.rs
```diff
@@ -126,11 +126,11 @@ where
 
     println!("\n");
     println!("🚀 {}, spawning.... with command:", node.name);
-    println!("{cmd} {}", args.join(" "));
+    println!("{program} {}", args.join(" "));
 
     let spawn_ops = SpawnNodeOptions {
         name: node.name.clone(),
-        command: cmd,
+        program,
```
pepoviola, line 133: I know that naming is hard, but what are the reason behind calling this `program`?

crates/provider/src/native.rs
```diff
@@ -567,10 +625,10 @@ fn retrieve_pid_from_process(process: &Child, node_name: &str) -> Result<Pid, Pr
     ))
 }
 
-fn create_stream_polling_task(
-    stream: impl AsyncRead + Unpin + Send + 'static,
-    tx: Sender<Result<Vec<u8>, Error>>,
-) -> JoinHandle<()> {
+fn create_stream_polling_task<S>(stream: S, tx: Sender<Result<Vec<u8>, Error>>) -> JoinHandle<()>
+where
```
pepoviola, line 629: Nice :)

crates/orchestrator/src/spawner.rs
l0r1s: I used what is used in [std::process::Command::new](https://doc.rust-lang.org/std/process/struct.Command.html#method.new) and [tokio::process::Command::new](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.new), I think ```program``` is overall a better name than ```command``` because it could either be an executable path (like a script or full binary path) or an executable (present in the PATH or similar), a command fits more for shell built-in commands imho

crates/provider/src/native.rs
```diff
@@ -593,44 +651,63 @@ fn create_stream_polling_task(
...
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .kill_on_drop(true)
-        .spawn()
+    let process = process_manager
+        .spawn(
+            Command::new(program)
+                .args(args)
+                .envs(env.to_owned())
+                .args(args)
```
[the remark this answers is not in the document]
l0r1s: Good catch !

crates/support/src/process.rs
```diff
@@ -0,0 +1,119 @@
...
+        self.args = args
+            .into_iter()
+            .map(|arg| arg.as_ref().to_os_string())
+            .collect();
+        self
+    }
+
+    pub fn envs<I, K, V>(mut self, vars: I) -> Self
+    where
+        I: IntoIterator<Item = (K, V)>,
```
pepoviola, line 55: nice :)

pepoviola (approved):

PR #123, opened 2023-10-19
Refacto/native provider

l0r1s:
- Split native provider into multiple smaller files and each with their respective tests
- Update ProviderNode.command() to ProviderNode.program()
- Update ProviderNamespace.id() to ProviderNamespace.name()


crates/orchestrator/src/network.rs
```diff
@@ -215,8 +215,8 @@ impl<T: FileSystem> Network<T> {
         self.parachains.insert(para.para_id, para);
     }
 
-    pub(crate) fn id(&self) -> &str {
-        self.ns.id()
+    pub(crate) fn name(&self) -> &str {
```
wirednkod, line 218:
```suggestion
    pub(crate) fn name_space(&self) -> &str {
```

pepoviola (approved):

PR #124, opened 2023-10-20
Add Orchestrator::native()

pgherveou:
Hey there just looking at your examples.
I thought that this PR could help simplify things a bit and avoid exposing internals struct to consumers of the sdk


pgherveou:
Also not sure if this is on your roadmap. But it would be nice if you could integrate with subxt 
so that you can easily create rpc client for the spawned network

Something along these lines 
```rust
    /// ...
    let mut network = Orchestrator::native().spawn(config).await?;

    // create a rpc client for the relay
    let relay_client = network.rpc_client(RelayChain("alice")).expect("Alice should be running");
    
    // create a rpc client for the parachain
    let collator_client = network.rpc_client(Parachain{ id: 100, name: "collator1"}).expect("Collator should be running");
```

crates/examples/examples/simple_network_example.rs
```diff
@@ -2,19 +2,13 @@
...
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let config = NetworkConfig::load_from_toml("./crates/examples/examples/0001-simple.toml")
         .expect("errored?");
 
-    let fs = LocalFileSystem;
-    let pm = OsProcessManager;
-    let provider = NativeProvider::new(fs.clone(), pm);
-    let orchestrator = Orchestrator::new(fs, provider);
-    orchestrator.spawn(config).await?;
+    Orchestrator::native().spawn(config).await?;
```
wirednkod, line 11: How does this uses of the LocalFileSystem?

wirednkod (commented):
This wont work, as the change

crates/examples/examples/simple_network_example.rs
pgherveou: I think Rust can infer the generics since it knows that native is implemented for T = LocalFileSystem

pgherveou:
Ping @wirednkod @pepoviola 
This is another quick one lmk what you think or if we should close it

wirednkod (approved):
> Ping @wirednkod @pepoviola This is another quick one lmk what you think or if we should close it

I fixed a conflict - if tests pass then we will be good to go

pepoviola:
Hi @pgherveou, thanks for the contribution but in this case we want to have some higher level entity that abstract those internals and expose this nice methods (e.g native, k8s, etc) but not as part of the orchestrator since that couple the orchestrator to a particular provider and we want to keep those separate. I'm not sure how this `module` will be called (naming is hard :) ), but we definitely wants to expose simple methods like the one you propose here.

We can work on the new mod here if you want, 
Thx!!

pgherveou:
Sounds good 👍 closing here for now

PR #125, opened 2023-10-20
Expose subxt onclineclient / rpc

pgherveou:
just wanted to throw a quick implementation that exposes the OnclineClient on the node.
This would make the sdk super useful to enable cross chain e2e tests in ink!

### DIFF
--- a/crates/orchestrator/src/network/node.rs
+++ b/crates/orchestrator/src/network/node.rs
@@ -3,6 +3,7 @@ use std::{sync::Arc, time::Duration};
 use anyhow::anyhow;
 use prom_metrics_parser::MetricMap;
 use provider::DynNode;
+use subxt::{backend::rpc::RpcClient, OnlineClient, PolkadotConfig};
 use tokio::sync::RwLock;
 
 use crate::network_spec::node::NodeSpec;
@@ -15,6 +16,8 @@ pub struct NetworkNode {
     pub(crate) spec: NodeSpec,
     pub(crate) name: String,
     pub(crate) ws_uri: String,
+    rpc: RpcClient,
+    client: OnlineClient<PolkadotConfig>,
     pub(crate) prometheus_uri: String,
     metrics_cache: Arc<RwLock<MetricMap>>,
 }
@@ -24,6 +27,8 @@ impl NetworkNode {
     pub(crate) fn new<T: Into<String>>(
         name: T,
         ws_uri: T,
+        rpc: RpcClient,
+        client: OnlineClient<PolkadotConfig>,
         prometheus_uri: T,
         spec: NodeSpec,
         inner: DynNode,
@@ -31,6 +36,8 @@ impl NetworkNode {
         Self {
             name: name.into(),
             ws_uri: ws_uri.into(),
+            rpc,
+            client,
             prometheus_uri: prometheus_uri.into(),
             inner,
             spec,
@@ -43,6 +50,14 @@ impl NetworkNode {
     pub async fn pause(&self) -> Result<(), anyhow::Error> {
         self.inner.pause().await?;
         Ok(())
+    }
+
+    pub fn rpc(&self) -> RpcClient {
+        self.rpc.clone()
+    }
+
+    pub fn client(&self) -> OnlineClient<PolkadotConfig> {
+        self.client.clone()
     }
 
     /// Resume the node, this is implemented by resuming the
--- a/crates/orchestrator/src/spawner.rs
+++ b/crates/orchestrator/src/spawner.rs
@@ -1,10 +1,13 @@
 use std::path::PathBuf;
 
+use anyhow::Context;
+use futures::Future;
 use provider::{
     constants::LOCALHOST,
     types::{SpawnNodeOptions, TransferedFile},
     DynNamespace,
 };
+use subxt::{backend::rpc::RpcClient, OnlineClient, PolkadotConfig};
 use support::fs::FileSystem;
 
 use crate::{
@@ -149,6 +152,7 @@ where
     let running_node = ctx.ns.spawn_node(spawn_ops).await?;
 
     let ws_uri = format!("ws://{}:{}", LOCALHOST, node.rpc_port.0);
+
     let prometheus_uri = format!("http://{}:{}/metrics", LOCALHOST, node.prometheus_port.0);
     println!("🚀 {}, should be running now", node.name);
     println!(
@@ -158,11 +162,43 @@ where
     println!("🚀 {} : metrics link {prometheus_uri}", node.name);
     println!("📓 logs cmd: tail -f {}/{}.log", base_dir, node.name);
     println!("\n");
+
+    let client = retry(|| async {
+        OnlineClient::from_url(&ws_uri)
+            .await
+            .context(format!("Failed to connect to node rpc at {ws_uri}"))
+    })
+    .await?;
+
+    let rpc = RpcClient::from_url(&ws_uri)
+        .await
+        .context(format!("Failed to connect to rpc client at {ws_uri}"))?;
+
     Ok(NetworkNode::new(
         node.name.clone(),
         ws_uri,
+        rpc,
+        client,
         prometheus_uri,
         node.clone(),
         running_node,
     ))
+}
+
+async fn retry<T, F>(connect: F) -> anyhow::Result<OnlineClient<PolkadotConfig>>
+where
+    T: Future<Output = anyhow::Result<OnlineClient<PolkadotConfig>>>,
+    F: Fn() -> T,
+{
+    let mut retries = 5;
+    loop {
+        match connect().await {
+            Err(_) if retries >= 0 => {
+                println!("Error connecting, retrying ...");
+                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                retries -= 1;
+            },
+            res => break res,
+        }
+    }
 }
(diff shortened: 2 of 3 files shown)


pepoviola:
Hey @pgherveou thanks for your contribution 🙌, something like this is definitely useful for teams but we need to brainstorming about the `API`  since we also want to expose other ways to interact with the node in the context of the network.

pepoviola:
> Sure feel free to close, just wanted to try it out

No, I don't think we should close it. I think is in the right path and worth to explore it and see the feedback from other teams. Then we can rethink some of the abstractions. @wirednkod / @l0r1s .

Just FYI @pgherveou we want to add also support to run `embedded` pjs scripts like the `javascript` tab of poladot.js inside the `Network` instance.

crates/orchestrator/src/spawner.rs
```diff
@@ -158,11 +162,40 @@ where
     println!("🚀 {} : metrics link {prometheus_uri}", node.name);
     println!("📓 logs cmd: tail -f {}/{}.log", base_dir, node.name);
     println!("\n");
+
+    let client = retry(|| async { OnlineClient::from_url(&ws_uri).await })
+        .await
+        .context(format!("Failed to connect to node rpc at {ws_uri}"))?;
```
wirednkod, line 168: should be added to the `crates/configuration/src/shared/constants.rs`

crates/orchestrator/src/spawner.rs
```diff
@@ -158,11 +162,40 @@ where
...
+
+    let rpc = RpcClient::from_url(&ws_uri)
+        .await
+        .context(format!("Failed to connect to rpc client at {ws_uri}"))?;
```
wirednkod, line 172: should be added to the `crates/configuration/src/shared/constants.rs`

crates/orchestrator/src/spawner.rs
```diff
@@ -158,11 +162,40 @@ where
...
+async fn retry<T, F, R, E>(connect: F) -> Result<R, E>
+where
+    T: Future<Output = Result<R, E>>,
+    F: Fn() -> T,
+{
+    let mut retries = 5;
+    loop {
+        match connect().await {
+            Err(_) if retries >= 0 => {
+                println!("Error connecting, retrying ...");
```
wirednkod, line 194: should be added to the `crates/configuration/src/shared/constants.rs`

crates/orchestrator/src/spawner.rs
```diff
@@ -158,11 +162,40 @@ where
...
         running_node,
     ))
 }
+
+async fn retry<T, F, R, E>(connect: F) -> Result<R, E>
+where
+    T: Future<Output = Result<R, E>>,
+    F: Fn() -> T,
+{
+    let mut retries = 5;
```
wirednkod, line 190:
```suggestion
    let mut retries = 3;
```

pgherveou:
> Just FYI @pgherveou we want to add also support to run embedded pjs scripts like the javascript tab of poladot.js inside the Network instance.

Should be nice, but not really useful for our usecase. We mainly want to interact with the node through subxt. That's what we do today, but with a solochain

pepoviola:
> > Just FYI @pgherveou we want to add also support to run embedded pjs scripts like the javascript tab of poladot.js inside the Network instance.
> 
> Should be nice, but not really useful for our usecase. We mainly want to interact with the node through subxt. That's what we do today, but with a solochain

Sounds good! and I think this is a great step forward :)

pgherveou:
This is the code I could replace with zsdk. 

https://github.com/paritytech/ink/blob/02c4f0d5981f498051a954bd743671310cb34f15/crates/e2e/src/node_proc.rs#L51

So pretty much what's missing are these 2 apis I added here. Other methods like kill are I believe already available. 

Then I would also need to specify a custom path for the collator, but I believe it can already be specified as well.

pepoviola:
Yes, looks like an awesome use-case. We were thinking on make the client/rpc call to create the client and not auto-connect the nodes at spawn phase. Something like from end-user perspective

```rs
let client = network.get_node("alice")?.client().await?;
```

crates/orchestrator/src/spawner.rs
pgherveou: Are you moving all &str into this file? This might make logging a bit more tedious, curious what you gain from this?

pgherveou:
One thing that would be nice as well is to use log and a custom target instead of println, so that these logs can easily be stripped from our tests.

crates/orchestrator/src/spawner.rs
pgherveou: when I run the example It fails between 3/4 times so 5 seems like a good default at least for now

pepoviola: This is mostly because the `auto-connect` logic is trying yo connect before the ws server is ready. If we change the logic to connect on-demand the number of retries will be less.
Thx!

crates/orchestrator/src/spawner.rs
```diff
@@ -158,11 +162,40 @@ where
...
+where
+    T: Future<Output = Result<R, E>>,
+    F: Fn() -> T,
+{
+    let mut failed_count = 0;
+    loop {
+        match connect().await {
+            Err(_) if retry_count > failed_count => {
+                failed_count += 1;
+                println!("Error connecting, retrying {failed_count}/{retry_count} ...");
```
pepoviola, line 195: Logging is on our `//TODO` list, we can add here a comment to easily update with when we start moving all those println to the logger impl.

pepoviola:
> > Yes, looks like an awesome use-case. We were thinking on make the client/rpc call to create the client and not auto-connect the nodes at spawn phase. Something like from end-user perspective
> > ```rust
> > let client = network.get_node("alice")?.client().await?;
> > ```
> 
> that would work too, as long as the sdk can handle the retry logic

@pgherveou, I'm not familiarized with the subxt internal, did you know if they handle reconnections? I wonder if make sense to `cache` the client in the `Node` as is in your `pr` (even if we change to connect on-demand).
Thx!

pgherveou:
> @pgherveou, I'm not familiarized with the subxt internals, did you know if they handle reconnections? I wonder if make sense to `cache` the client in the `Node` as is in your `pr` (even if we change to connect on-demand). Thx!

under the hood,  looks like it's a [tokio TcpStream](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html). if the Node is stopped you will have to recreate the client.

You will receive this kind of error if you try to read from it. 
```
RPC error: The background task been terminated because: Networking or low-level protocol error: WebSocket connection error: connection closed; restart required)
```

Maybe creating the client on demand like you are suggesting and letting the consumer deal with re-connection is good enough.
You can maybe just create a temporary one just to poll until the network is ready. There might be option to also pass a healthy tcpstream to build the client, if you want to keep a cached connection around.

pgherveou:
> @pgherveou, I'm not familiarized with the subxt internals, did you know if they handle reconnections? I wonder if make sense to `cache` the client in the `Node` as is in your `pr` (even if we change to connect on-demand). Thx!

updated the PR to do it lazily, this remove the need for the custom retry function as well

crates/orchestrator/src/spawner.rs
```diff
@@ -149,6 +150,7 @@ where
     let running_node = ctx.ns.spawn_node(spawn_ops).await?;
 
     let ws_uri = format!("ws://{}:{}", LOCALHOST, node.rpc_port.0);
+
```
pgherveou, line 153:
```suggestion
```

crates/orchestrator/src/spawner.rs
```diff
@@ -158,6 +160,7 @@ where
     println!("🚀 {} : metrics link {prometheus_uri}", node.name);
     println!("📓 logs cmd: tail -f {}/{}.log", base_dir, node.name);
     println!("\n");
+
```
pgherveou, line 163:
```suggestion
```

crates/orchestrator/src/spawner.rs
```diff
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use futures::Future;
```
pgherveou, line 3:
```suggestion
```

crates/orchestrator/src/network/node.rs
```diff
@@ -45,6 +46,14 @@ impl NetworkNode {
         Ok(())
     }
 
+    pub async fn rpc(&self) -> Result<RpcClient, subxt::Error> {
+        RpcClient::from_url(&self.ws_uri).await
+    }
+
+    pub async fn client(&self) -> Result<OnlineClient<PolkadotConfig>, subxt::Error> {
```
pepoviola, line 53: Does make sense to allow users to also specify the concrete `Config` type?

pgherveou: mmm good point. Using moonbeam for example you would need to specify a custom config.
A bit of a shame that Rust does not let you specify a default value for the generic trait here.

I mean you could but then rustc complains with
```
defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions  this was previously accepted by the compiler but is being phased out;
```

pepoviola:
Looks good @pgherveou! thanks for your contribution 🙌 🙌  I will fix the fmt/clippy and merge this one.
Again, thanks for your contribution!

wirednkod (approved):
Besides the clippy issues this make sense. Thank you @pgherveou

pgherveou:
🙏  looks like there are quite a lot of clippy issues not related to the changes here. Should you ship the fixes in a different PR?

pepoviola:
merged, thanks @pgherveou!!

PR #126, opened 2023-11-02
Expose Top level API through zombienet-sdk crate

pgherveou:
Following up on take 1 https://github.com/paritytech/zombienet-sdk/pull/124

Apart from better logs, I realize that one of the only thing missing now for us to start consuming this project  is a top level package available (with it's dependencies) on crates.io. This PR serves mainly as a discussion point to get there.
It would be great if we could publish an early version (0.0.1?) on crates.io so that we can start using this asap.

### DIFF
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -1,0 +1,20 @@
+use async_trait::async_trait;
+pub use configuration::{NetworkConfig, NetworkConfigBuilder, RegistrationStrategy};
+pub use orchestrator::{errors::OrchestratorError, network::Network, AddNodeOpts, Orchestrator};
+use provider::NativeProvider;
+use support::{fs::local::LocalFileSystem, process::os::OsProcessManager};
+
+#[async_trait]
+pub trait Spawner {
+    /// Spawns a network using the native provider.
+    async fn spawn_native(self) -> Result<Network<LocalFileSystem>, OrchestratorError>;
+}
+
+#[async_trait]
+impl Spawner for NetworkConfig {
+    async fn spawn_native(self) -> Result<Network<LocalFileSystem>, OrchestratorError> {
+        let provider = NativeProvider::new(LocalFileSystem {}, OsProcessManager {});
+        let orchestrator = Orchestrator::new(LocalFileSystem {}, provider);
+        orchestrator.spawn(self).await
+    }
+}
--- a/crates/sdk/src/lib.rs
+++ b/crates/sdk/src/lib.rs
@@ -5,13 +5,13 @@ use provider::NativeProvider;
 use support::{fs::local::LocalFileSystem, process::os::OsProcessManager};
 
 #[async_trait]
-pub trait Spawner {
+pub trait NetworkConfigExt {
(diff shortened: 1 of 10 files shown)


crates/sdk/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+use async_trait::async_trait;
+pub use configuration::{NetworkConfig, NetworkConfigBuilder, RegistrationStrategy};
+pub use orchestrator::{errors::OrchestratorError, network::Network, AddNodeOpts, Orchestrator};
+use provider::NativeProvider;
+use support::{fs::local::LocalFileSystem, process::os::OsProcessManager};
+
+#[async_trait]
+pub trait Spawner {
```
pepoviola, line 8: I think we can rethink this `naming`, something like `NetworkConfigSpawnExt` or just `NetworkConfigExt`. 
@wirednkod / @l0r1s / @pgherveou wdyt?

l0r1s:
Hi @pgherveou, sure ! (and some docs probably)

crates/sdk/src/lib.rs
l0r1s: Yes, we can do this temporarily, not a issue to change this in the future :+1:

pgherveou: 👍  NetworkConfigExt sounds a bit better

pgherveou:
@pepoviola most urgent missing bit, apart from this top level package is fixing this todo! that prevent us from defining a custom command.

https://github.com/paritytech/zombienet-sdk/blob/m/crates/orchestrator/src/generators/chain_spec.rs#L364

The other thing is: `get_node` only return validator node for now, should we adapt the code like this so that we can get reference to any node, including parachain collators?

`add_para` probably need to be adjusted like this 
```diff
     pub(crate) fn add_para(&mut self, para: Parachain) {
+        para.collators.iter().for_each(|collator| {
+            self.nodes_by_name
+                .insert(collator.name.clone(), collator.clone());
+        });
         self.parachains.insert(para.para_id, para);
     }
```

pepoviola:
> @pepoviola most urgent missing bit, apart from this top level package is fixing this todo! that prevent us from defining a custom command.
> 
> https://github.com/paritytech/zombienet-sdk/blob/m/crates/orchestrator/src/generators/chain_spec.rs#L364
> 
> The other thing is: `get_node` only return validator node for now, should we adapt the code like this so that we can get reference to any node, including parachain collators?
> 
> `add_para` probably need to be adjusted like this
> 
> ```diff
>      pub(crate) fn add_para(&mut self, para: Parachain) {
> +        para.collators.iter().for_each(|collator| {
> +            self.nodes_by_name
> +                .insert(collator.name.clone(), collator.clone());
> +        });
>          self.parachains.insert(para.para_id, para);
>      }
> ```

Thanks @pgherveou, I just open this https://github.com/paritytech/zombienet-sdk/issues/127 to make all the node available through `get_node`. Related to the chain_spec, did you need to use a custom command to generate it?

Thx!

pepoviola (approved):

Issue #35, opened 2023-05-18
Kubernetes (`./crates/providers/kubernetes` - features = "kubernetes")


wirednkod:

pepoviola:
K8s provider is now merged in main. Closing here.

PR #130, opened 2023-11-07
Feat/k8s provider

l0r1s:
This PR introduces the Kubernetes Provider for the SDK, with the use of kube.rs for common operation and a custom file server using axum to transfer files back and forth the nodes.

https://github.com/paritytech/zombienet-sdk/issues/35

This also includes the refacto of the native provider to reflect changes to the provider structure with weak references and simplifying the use of abstractions


crates/provider/src/kubernetes/client/kube_rs.rs
```diff
@@ -0,0 +1,379 @@
...
+    filesystem: FS,
+}
+
+impl<FS> KubeRsKubernetesClient<FS>
+where
+    FS: FileSystem + Send + Sync + Clone,
+{
+    async fn new(filesystem: FS) -> Result<Self, Box<dyn std::error::Error>> {
+        Ok(Self {
+            // TODO: make it more flexible with path to kube config
```
pepoviola, line 39: I think is ok for now, I saw most of the time usage of the default location of the config.

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -0,0 +1,239 @@
...
+            .client
+            .create_pod_logs_stream(&options.name, &self.name)
+            .await
+            .unwrap();
+
+        // handle log writing
+        let (logs_tx, rx) = mpsc::channel(10);
+        let log_reading_handle = create_stream_polling_task(logs_stream, logs_tx);
+        let log_writing_handle =
+            create_log_writing_task(rx, self.filesystem.clone(), log_path.clone());
```
pepoviola, line 184: I don't think we want to write the logs locally, we can always retrieve from the pods (or from grafana if we relay with loki).

crates/provider/src/kubernetes/node.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub(super) inner: Arc<RwLock<KubernetesNodeInner>>,
+    pub(super) filesystem: FS,
+    pub(super) client: KC,
+    pub(super) namespace: WeakKubernetesNamespace<FS, KC>,
+}
+
+pub(super) struct KubernetesNodeInner {
+    pub(super) log_reading_handle: JoinHandle<()>,
+    pub(super) log_writing_handle: JoinHandle<()>,
+}
```
pepoviola, line 40: Not needed also.

crates/provider/src/kubernetes/zombie-wrapper.sh
```diff
@@ -0,0 +1,116 @@
...
+
+# init empty
+child_pid=""
+
+# get the command to exec
+CMD=($@)
+
+echo "COMMAND TO RUN IS: $CMD"
+
+start() {
```
pepoviola, line 54: nice!

crates/provider/src/shared/types.rs
```diff
@@ -241,8 +249,17 @@ impl TransferedFile {
         Self {
             local_path: local_path.as_ref().into(),
             remote_path: remote_path.as_ref().into(),
+            mode: "0644".to_string() // default to rw-r--r--
```
pepoviola, line 252: 👍

crates/provider/src/kubernetes/client/kube_rs.rs
```diff
@@ -0,0 +1,379 @@
...
+            })?;
+        let actual_hash = sha256sum_stdout
+            .split_whitespace()
+            .next()
+            .expect("sha256sum output should be in the form `hash<spaces>filename`");
+
+        // get the hash of the file
+        let expected_hash = hex::encode(sha2::Sha256::digest(&contents));
+        if actual_hash != expected_hash {
+            // TODO: should we delete partially copied file here?
```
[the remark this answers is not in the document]
l0r1s: Well, actually I've made a few tests and it doesn't report error if the file is not completely written, so I made this as in the [example present here](https://github.com/kube-rs/kube/blob/main/examples/pod_cp.rs) where they verify the file is written (here I just check the checksum and it doesn't add much overhead)

crates/provider/src/kubernetes/client/kube_rs.rs
l0r1s: Yes, was just an idea for a specific use case, can be let here and in the future if we have time to make things cleaner, it's more a "idea list"

crates/provider/src/shared/types.rs
```diff
@@ -231,6 +267,8 @@ impl RunScriptOptions {
 pub struct TransferedFile {
     pub local_path: PathBuf,
     pub remote_path: PathBuf,
+    // TODO: Can be narrowed to have strict typing on this?
```
pepoviola, line 270: Maybe using https://github.com/sourcefrog/unix_mode ?

pepoviola:
ping @l0r1s, let me know if I can resolve this conflicts and review this one.
Thx!

crates/provider/src/shared/types.rs
l0r1s: The issue is that it only allow to converts a mode as string, no the other way and I think that's what we would expect, user would pass string input and we convert it into a mode ? Maybe a custom type ?

crates/provider/src/kubernetes/namespace.rs
l0r1s: Removed, make code cleaner, k8s is complex enough without it, can still be used in the future if we need it (from git history)

l0r1s:
> ping @l0r1s, let me know if I can resolve this conflicts and review this one. Thx!

I think we are good

crates/provider/src/kubernetes/node.rs
```diff
@@ -43,9 +43,14 @@ where
 {
     pub(super) async fn new(
         namespace: &Weak<KubernetesNamespace<FS>>,
+        namespace_base_dir: &PathBuf,
```
pepoviola, line 46: Can we get this field from the weak ref of the namespace?

crates/orchestrator/src/spawner.rs
```diff
@@ -174,14 +176,31 @@ where
...
+        ])
+        .await?;
+
+        (
+            ports[0].unwrap_or(node.ws_port.0),
+            ports[0].unwrap_or(node.prometheus_port.0),
+        )
+    } else {
+        (node.ws_port.0, node.prometheus_port.0)
+    };
```
l0r1s, line 193: I think the "default" is a bit redundant (```node.ws_port.0``` and  ```node.prometheus_port.0```), maybe we could refacto to have the default initialized and then if not running in CI, creating the port forward, also the node.prometheus_port.0 should be assigne port[1] line 189:

```suggestion

    let (mut rpc_port_external, mut prometheus_port_external) =
        (node.ws_port.0, node.prometheus_port.0);

    if !running_in_ci() {
        let ports = futures::future::try_join_all([
            running_node.create_port_forward(node.ws_port.0, RPC_PORT),
            running_node.create_port_forward(node.prometheus_port.0, PROMETHEUS_PORT),
        ])
        .await?;

        (rpc_port_external, prometheus_port_external) = (ports[0].unwrap(), ports[1].unwrap());
    }
    ```

File: crates/provider/src/kubernetes/node.rs [hunk 10]
```diff
@@ -364,6 +367,30 @@ where
...
+        // If the fwd exist just return the local port
+        if let Some(fwd_info) = self.port_fwds.read().await.get(&remote_port) {
+            return Ok(Some(fwd_info.0));
+        };
+
+        let (port, task) = self
+            .k8s_client
+            .create_pod_port_forward(&self.namespace_name(), &self.name, local_port, remote_port)
+            .await
+            .map_err(|err| ProviderError::FileServerSetupError(err.into()))?;
```
l0r1s, line 384: I think we can extract a new error ```ProviderError::PortForwardError```

l0r1s (commented):
Small changes to make code cleaner and simplify debugging 👍

crates/provider/src/kubernetes/node.rs
pepoviola: Yes, sounds good 👍

crates/orchestrator/src/spawner.rs
pepoviola: Sounds good, I will change it. 👍

PR #133, opened 2023-11-08
Fix: paras spec file name

pepoviola:
Fix parachain spec filename (local) to use when we spawn collators. Also, adding `Context` to spawner to improve error reporting.

cc @pgherveou

### DIFF
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -289,11 +289,7 @@ where
                 // add the spec to global files to inject
                 let spec_name = chain_spec.chain_spec_name();
                 para_files_to_inject.push(TransferedFile {
-                    local_path: PathBuf::from(format!(
-                        "{}/{}.json",
-                        ns.base_dir().to_string_lossy(),
-                        spec_name
-                    )),
+                    local_path: ns.base_dir().join(format!("{}.json", spec_name)),
                     remote_path: PathBuf::from(format!("/cfg/{}.json", para.id)),
                 });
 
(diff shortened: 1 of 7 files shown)


crates/orchestrator/src/lib.rs
```diff
@@ -279,9 +279,24 @@ where
...
                 let id = chain_spec.read_chain_id(&scoped_fs).await?;
+
+                // add the spec to global files to inject
+                let spec_name = chain_spec.chain_spec_name();
+                para_files_to_inject.push(TransferedFile {
+                    local_path: PathBuf::from(format!(
+                        "{}/{}.json",
+                        ns.base_dir().to_string_lossy(),
+                        spec_name
+                    )),
```
pgherveou, line 296:
```suggestion
                    local_path: ns.base_dir().join(format!("{}.json", spec_name)),
```

crates/orchestrator/src/spawner.rs
```diff
@@ -150,7 +151,11 @@ where
     node.rpc_port.drop_listener();
     node.prometheus_port.drop_listener();
 
-    let running_node = ctx.ns.spawn_node(spawn_ops).await?;
+    let err_context = format!(
+        "Failed to spawn node: {} with opts: {:#?}",
+        node.name, spawn_ops
+    );
+    let running_node = ctx.ns.spawn_node(spawn_ops).await.context(err_context)?;
```
pgherveou, line 159: might be worth refactoring this spawn_node so it takes a ref option instead so that you can lazily compute the context here.

pgherveou (approved):
Tested it works! Thanks

crates/orchestrator/src/spawner.rs
pepoviola: Make sense, let me change it. Thx!

PR #150, opened 2024-01-18
Feat/file server

l0r1s:
- Added basic axum fileserver to upload/download files temporary to prevent error due to failure with exec + tar using kube.rs
- Added build and push to repository of image to be used in k8s


l0r1s:
I merge it to use it but still available for review @wirednkod @pepoviola

PR #154, opened 2024-01-24
Bump subxt

pgherveou:


pgherveou:
If we could publish the new alpha that would be great too 🙏

crates/orchestrator/src/network/parachain.rs
```diff
@@ -138,13 +138,26 @@ impl Parachain {
...
         // finalized block) to register the parachain
-        let result = api
+        let mut tx = api
             .tx()
             .sign_and_submit_then_watch_default(&sudo_call, &sudo)
             .await?;
 
-        let result = result.wait_for_in_block().await?;
-        info!("In block: {:#?}", result.block_hash());
+        while let Some(status) = tx.next().await {
```
pgherveou, line 146:
```suggestion
        // Below we use the low level API to replicate the `wait_for_in_block` behaviour
        // which was removed in subxt 0.33.0. See https://github.com/paritytech/subxt/pull/1237.
        while let Some(status) = tx.next().await {
```

pepoviola:
> If we could publish the new alpha that would be great too 🙏

Done, `v0.1.0-alpha.1` thanks!!

pgherveou:
Thanks Javier 🙏

PR #158, opened 2024-02-02
Feat/refacto native provider

l0r1s:
- Refacto of Native providerf to simplify the logic of cyclic references following emerging design from Kubernetes provider
- Removed flaky tests due to mocking of implementation details, tests are in progress to be integrated across the provider/orchestrator/configuration
- Removed unused code


crates/provider/src/native/node.rs
```diff
@@ -1,58 +1,259 @@
...
-    pub(super) config_dir: PathBuf,
-    pub(super) data_dir: PathBuf,
-    pub(super) relay_data_dir: PathBuf,
-    pub(super) scripts_dir: PathBuf,
-    pub(super) log_path: PathBuf,
-    pub(super) inner: Arc<RwLock<NativeNodeInner>>,
-    pub(super) filesystem: FS,
-    pub(super) process_manager: PM,
-    pub(super) namespace: WeakNativeNamespace<FS, PM>,
+    namespace: Weak<NativeNamespace<FS>>,
```
pepoviola, line 39: I'm now sure if we need to have a ref to the namespace in the node, in some cases I think we will need the `namespace name` but not the whole info.

crates/provider/src/native/provider.rs
```diff
@@ -5,52 +5,33 @@ use std::{
...
 };
 
-#[derive(Clone)]
-pub struct NativeProvider<FS, PM>
+pub struct NativeProvider<FS>
 where
     FS: FileSystem + Send + Sync + Clone,
-    PM: ProcessManager + Send + Sync + Clone,
 {
+    weak: Weak<NativeProvider<FS>>,
```
pepoviola, line 21: Let's touch base on this, I think we can simplify this.

PR #160, opened 2024-02-09
K8s wiring stuffs

pepoviola:


crates/orchestrator/src/shared/constants.rs
```diff
@@ -0,0 +1,8 @@
+/// Prometheus exporter default port
```
pepoviola, line 1: @l0r1s We should extract these to re-use in other crates. Maybe in support, wdyt?

crates/provider/src/kubernetes/scripts/zombie-wrapper.sh
```diff
@@ -20,7 +20,7 @@ fi
 echo "COMMANDS DEFINED"
 
 # add /cfg as first `looking dir` to allow to overrides commands.
-export PATH="{{REMOTE_DIR}}":$PATH
+export PATH="/cfg":$PATH
```
pepoviola, line 23: We are setting `/cfg` as remote dir in the pod fs, let's just use the same here.

crates/provider/src/shared/helpers.rs
```diff
@@ -0,0 +1,75 @@
+use std::collections::HashMap;
+
+use configuration::shared::constants::VALID_REGEX;
+use regex::{Captures, Regex};
+
+pub(crate) fn apply_replacements(text: &str, replacements: &HashMap<&str, &str>) -> String {
```
pepoviola, line 6: We can also support `tera` as templating lang (as we do in v1 with nunjucks) but for the moment is a little bit overkill.

crates/provider/src/shared/types.rs
```diff
@@ -11,7 +11,9 @@ pub type ExecutionResult = Result<String, (ExitStatus, String)>;
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct ProviderCapabilities {
+    /// Ensure that we have an image for each node (k8s/podman/docker)
```
pepoviola, line 14: We use this flag to decide other things, so maybe we need to rename it.

crates/orchestrator/src/generators/bootnode_addr.rs
```diff
@@ -1,13 +1,12 @@
-use std::{fmt::Display, net::IpAddr};
+use std::fmt::Display;
 
 use super::errors::GeneratorError;
 
 pub fn generate<T: AsRef<str> + Display>(
     peer_id: &str,
-    ip: &IpAddr,
+    ip: &str,
```
l0r1s, line 7: Why do we move from IpAddr to str ? We lose the type safety of ensuring the IpAddr is valid, wdyt ?

crates/orchestrator/src/generators/bootnode_addr.rs
```diff
@@ -83,7 +87,13 @@ mod tests {
             .iter()
             .map(|x| x.to_string())
             .collect();
-        let bootnode_addr = generate(peer_id, &LOCALHOST, 5678, args.iter().as_ref(), &None);
+        let bootnode_addr = generate(
+            peer_id,
+            &LOCALHOST.to_string(),
+            5678,
```
l0r1s, line 93: Maybe in the future we could create something like ```type Host = (IpAddr::V4, u16)```

crates/orchestrator/src/generators/command.rs
```diff
@@ -9,6 +9,7 @@ pub struct GenCmdOptions<'a> {
     pub relay_data_path: &'a str,
     pub use_wrapper: bool,
     pub bootnode_addr: Vec<String>,
+    pub use_default_ports_in_cmd: bool,
```
l0r1s, line 12: I don't really understand what it means, aren't we suppose to fallback on default port if no port are provided anyway ?

crates/orchestrator/src/generators/command.rs
```diff
@@ -244,12 +246,30 @@ pub fn generate_for_node(
...
     // ports
+
+    // Prometheus
     tmp_args.push("--prometheus-port".into());
-    tmp_args.push(node.prometheus_port.0.to_string());
+    tmp_args.push(if options.use_default_ports_in_cmd {
+        PROMETHEUS_PORT.to_string()
+    } else {
+        node.prometheus_port.0.to_string()
+    });
```
l0r1s, line 256: I think we could directly set the node.prometheus_port as a default when when initialize, and override when user set it ?

crates/orchestrator/src/lib.rs
```diff
@@ -224,12 +225,17 @@ where
...
                 generators::generate_node_bootnode_addr(
                     &node.spec.peer_id,
-                    &LOCALHOST,
-                    node.spec.p2p_port.0,
+                    &ip,
+                    if ctx.ns.capabilities().requires_image {
+                        P2P_PORT
+                    } else {
+                        node.spec.p2p_port.0
+                    },
```
l0r1s, line 238: Shouldnt be ```options.use_default_ports_in_cmd``` instead?

crates/orchestrator/src/shared/constants.rs
l0r1s: Yes sure

crates/orchestrator/src/spawner.rs
```diff
@@ -116,6 +116,12 @@ where
         relay_data_path: &relay_data_path, // TODO: get from provider
         use_wrapper: false,                // TODO: get from provider
         bootnode_addr: ctx.bootnodes_addr.clone(),
+        // IFF the provider require an image (e.g k8s) we should use the default ports in the cmd.
+        use_default_ports_in_cmd: if ctx.ns.capabilities().requires_image {
+            true
+        } else {
+            false
+        },
```
l0r1s, line 124: ```requires_image``` is already a boolean:

```suggestion
        use_default_ports_in_cmd: ctx.ns.capabilities().requires_image,
```

crates/provider/src/kubernetes/client.rs
```diff
@@ -145,7 +148,7 @@ impl KubernetesClient {
             .await
             .map_err(|err| Error::from(anyhow!("error while creating pod {name}: {err}")))?;
 
-        await_condition(pods, name, conditions::is_pod_running())
+        await_condition(pods, name, helpers::is_pod_ready())
```
l0r1s, line 151: Why this modification ? ```is_pod_running()``` doesn't works ?

crates/provider/src/kubernetes/client.rs
```diff
@@ -369,6 +386,40 @@ impl KubernetesClient {
...
+
+        api.create(
+            &PostParams::default(),
+            &serde_yaml::from_str(raw_manifest).unwrap(),
+        )
+        .await
+        .map_err(|err| {
+            Error::from(anyhow!(
+                "error while creating static-config {raw_manifest}: {err}"
+            ))
```
l0r1s, line 417: Maybe we should write the manifest as Rust, this way, we benefit from typechecking and now if something doesn't works between kubernetes version upgrade due to API changes:

[LimitRange](https://docs.rs/k8s-openapi/latest/k8s_openapi/api/core/v1/struct.LimitRange.html)
[NetworkPolicy](https://docs.rs/k8s-openapi/latest/k8s_openapi/api/networking/v1/struct.NetworkPolicy.html)

Wdyt ?

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -100,7 +107,21 @@ where
...
+            ),
+            (
+                "projectName".to_string(),
+                env::var("CI_PROJECT_NAME").unwrap_or("".to_string()),
+            ),
+            (
+                "projectId".to_string(),
+                env::var("CI_PROJECT_ID").unwrap_or("".to_string()),
+            ),
+        ]);
```
l0r1s, line 124: These labels doesn't need to be namespaced ? (like kubernetes.io/app=somevalue ...)

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -121,15 +142,44 @@ where
...
+            self.k8s_client
+                .create_static_resource(
+                    &self.name,
+                    include_str!("./static-configs/baseline-resources.yaml"),
+                )
+                .await
+                .map_err(|err| {
+                    ProviderError::CreateNamespaceFailed(self.name.to_string(), err.into())
+                })?;
+        }
```
l0r1s, line 170: Here, from what I understand, we deploy static resources if we are running in CI, shouldn't be the contrary ? To not deploy them in CI ?

Also, we can extract a function ```initialize_static_resources``` to avoid cluttering the ```initialize_k8s```:


Something like this:
```suggestion
        if helpers::running_in_ci() {
            self.initialize_static_resources().await?;
        }
```

```rust
mod helpers {
    pub fn running_in_ci() -> bool {
        env::var("RUN_IN_CI").unwrap_or_default() == "1"
    }
}
```

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -141,8 +191,20 @@ where
...
+                    http_get: Some(HTTPGetAction {
+                        path: Some("/".to_string()),
+                        port: IntOrString::Int(80),
+                        ..Default::default()
+                    }),
+                    initial_delay_seconds: Some(5),
+                    period_seconds: Some(3),
+                    failure_threshold: Some(15),
+                    ..Default::default()
+                }),
```
l0r1s, line 204: 👍

crates/provider/src/kubernetes/node.rs
```diff
@@ -410,6 +475,26 @@ where
...
+        let status = self
+            .k8s_client
+            .pod_status(&self.namespace_name(), &self.name)
+            .await
+            .map_err(|_err| {
+                ProviderError::PauseNodeFailed(
+                    self.name.to_string(),
+                    anyhow!("error when pausing node: status"),
+                )
+            })?;
```
l0r1s, line 488: Probably not the right error here

crates/provider/src/shared/helpers.rs
l0r1s: I don't understand the purpose of this function, is it to update a running k8s resource ?

crates/provider/src/shared/types.rs
l0r1s: Yes, I saw above, maybe we could split this in multiple flags to avoid confusion and keep one flag for one thing

l0r1s (changes requested):

crates/orchestrator/src/generators/bootnode_addr.rs
pepoviola: Sounds good, let me rollback the change and add a parsing logic to get an `IpAddr` from String.

crates/orchestrator/src/generators/bootnode_addr.rs
pepoviola: I think keep addr/port as two args is ok, but we can re-visit in the future.

crates/orchestrator/src/generators/command.rs
pepoviola: Not really, because we `always` set the ports (randomly is the user don't set). So, at this point we will always have port assigned and that why we need to branch to expose the ports that the `image` are exposing. Those will be later using for port-forwarding to the local (randomly selected) ports.
make sense?

crates/orchestrator/src/lib.rs
pepoviola: We don't have access to those options here, is the same `if` we use to set the field.

crates/provider/src/kubernetes/client.rs
pepoviola: If the pod have some `probing` (like the one I added to the file server) k8s reports running and then (ones is verified) the `Ready` state. I think checking `ready` give us the possibility to use the built-in probes machinery. For pods not setting any probes the ready event is set at the same time that the running.

crates/provider/src/kubernetes/namespace.rs
pepoviola: No, we don't need to use namespace here since those are only used internally.

crates/provider/src/kubernetes/namespace.rs
pepoviola: No, we need to deploy those manifest only in `CI` to ensure `namespace` isolation (this could be deployed always) and the baseline resources should be deployed only for `CI` to ensure the minimal resource allocation.

crates/provider/src/kubernetes/client.rs
pepoviola: I think keep those manifest as `yaml` is ok, since most of the time are very static and if we need to change something is easily to review with @emamihe

crates/provider/src/shared/helpers.rs
pepoviola: No, the idea behind this `fn` is to allow us to have manifest (`yaml` files) that we need to inject values (e.g namespace or others) when we want to deploy them, so we can just past the _template_ and inject the values. This will also be handy to manage some info that is generated by the orchestrator and needs to be injected later in the command or arguments.

crates/provider/src/kubernetes/node.rs
pepoviola: Yes, nice catch :)

pepoviola:
Hi @l0r1s, I resolved the comments we talked. I will merge this one.
Thx!

PR #163, opened 2024-02-15
chore: gitlab runner test for new repo

emamihe:
This is a very simple gitlab CI manifest to just check the tag on this pipeline schedule it on Zombienet k8s runner and is the SA inject to the runners in privileged enough to run command within the cluster or not

### DIFF
--- a/.gitlab-ci.yaml
+++ b/.gitlab-ci.yaml
@@ -1,0 +1,10 @@
+stages:
+  - integration-test
+
+polkadot-integration-test:
+  stage: integration-test
+  image: paritytech/zombienet:v1.3.91
+  tags:
+    - zombienet-polkadot-integration-test
+  script:
+    - kubectl get ns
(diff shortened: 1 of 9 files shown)


.gitlab-ci.yaml
```diff
@@ -0,0 +1,12 @@
+stages:
+  - integration-test
+
+polkadot-integration-test:
+  stage: integration-test
+  image: paritytech/zombienet:v1.3.91
```
pepoviola, line 6: We need and image with rust to run the tests, I think we can use `paritytech/ci-unified` .

.gitlab-ci.yaml
```diff
@@ -0,0 +1,12 @@
+stages:
+  - integration-test
+
+polkadot-integration-test:
+  stage: integration-test
+  image: paritytech/ci-unified:latest
+  tags:
+    - zombienet-polkadot-integration-test
+  script:
+    - kubectl get ns
```
pepoviola, line 10: I don't think `ci-unified` have `kubectl` installed, we also don't need it. Maybe I can write an small script to test the access to the cluster easily.
Thx!

PR #170, opened 2024-02-23
feat: Allow add para to running network without register

pepoviola:
- Add `custom_parchain_fs_prefix` parameter to `add_para` fn (on running network), this allow to manage the fs layout and allow to add paras with the same id. 
- Minor clippy auto-fixes

cc: @metricaez

### DIFF
--- a/crates/orchestrator/src/network.rs
+++ b/crates/orchestrator/src/network.rs
@@ -312,7 +312,11 @@ impl<T: FileSystem> Network<T> {
 
     /// Add a new parachain to the running network
     ///
-    /// NOTE: para_id must be unique in the whole network.
+    /// # Arguments
+    /// * `para_config` - Parachain configuration to deploy
+    /// * `custom_relaychain_spec` - Optional path to a custom relaychain spec to use
+    /// * `custom_parchain_fs_prefix` - Optional prefix to use when artifacts are created
+    ///
     ///
     /// # Example:
     /// ```rust
@@ -334,7 +338,7 @@ impl<T: FileSystem> Network<T> {
     ///     .build()
     ///     .map_err(|_e| anyhow!("Building config"))?;
     ///
-    /// network.add_parachain(&para_config, None).await?;
+    /// network.add_parachain(&para_config, None, None).await?;
     ///
     /// #   Ok(())
     /// # }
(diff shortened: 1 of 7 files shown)


crates/provider/src/kubernetes/node.rs
```diff
@@ -73,7 +73,7 @@ where
         let filesystem = options.filesystem.clone();
 
         let base_dir =
-            PathBuf::from_iter([&options.namespace_base_dir, &PathBuf::from(options.name)]);
+            PathBuf::from_iter([(options.namespace_base_dir), &PathBuf::from(options.name)]);
```
l0r1s, line 76:
```suggestion
            PathBuf::from_iter([options.namespace_base_dir, &PathBuf::from(options.name)]);
```

crates/orchestrator/src/network.rs
```diff
@@ -375,13 +380,21 @@ impl<T: FileSystem> Network<T> {
...
             .build_chain_spec(&relay_chain_id, &self.ns, &scoped_fs)
             .await?;
-        scoped_fs.create_dir(para_spec.id.to_string()).await?;
+
+        // Para artifacts
+        let para_path_prefix = if let Some(custom_prefix) = custom_parchain_fs_prefix {
+            custom_prefix
+        } else {
+            para_spec.id.to_string()
+        };
```
l0r1s, line 389:
```suggestion
        let para_path_prefix = custom_parchain_fs_prefix.unwrap_or_else(|| para_spec.id.to_string());
```

crates/orchestrator/src/network.rs
```diff
@@ -312,7 +312,11 @@ impl<T: FileSystem> Network<T> {
 
     /// Add a new parachain to the running network
     ///
-    /// NOTE: para_id must be unique in the whole network.
+    /// # Arguments
+    /// * `para_config` - Parachain configuration to deploy
+    /// * `custom_relaychain_spec` - Optional path to a custom relaychain spec to use
+    /// * `custom_parchain_fs_prefix` - Optional prefix to use when artifacts are created
```
l0r1s, line 318: What does it means, isn't this leaking the way we add a parachain ? I mean, the outside world does needs to know about artifacts ?

l0r1s (changes requested):

crates/provider/src/kubernetes/node.rs
pepoviola: This one was fixed by clippy🤔

crates/orchestrator/src/network.rs
pepoviola: Not sure about this, I like `if let ..` but I'm open to change to `unwrap_or_else` if there is a good argument.

pepoviola:
Hi @l0r1s, I will merge this one as is now. We can touch base later if you want.
Thx!

PR #171, opened 2024-02-24
Fix builder pattern to allow real closures

s0me0ne-unkn0wn:
Sticking to the `fn` type in the builder pattern doesn't make much sense to me. Consider the following snippet:

```rust
    let overrides = json!({ "runtimeGenesis": { "patch": { "configuration": { "config": { "max_pov_size": 5_242_880 } } } } });
    let config = NetworkConfigBuilder::new()
        .with_relaychain(|r| {
            r.with_chain("rococo-local")
                .with_genesis_overrides(overrides)
                .with_node(|node| node.with_name("alice").with_command("polkadot"))
        })
        .build();
```

That wouldn't compile, as the `.with_relaychain()` argument closure captures the `overrides` variable and thus cannot be coerced to `fn` anymore. That is, `.with_genesis_overrides` becomes useless unless you want to pass a simple literal there.

On the other hand, all those closures are single-shot, allowing us to make them `FnOnce,` which is relatively relaxed. If `.with_relaychain()` accepts `impl FnOnce` instead of `fn`, the example above compiles successfully.

### DIFF
--- a/crates/configuration/src/network.rs
+++ b/crates/configuration/src/network.rs
@@ -324,7 +324,7 @@ impl NetworkConfigBuilder<Initial> {
     /// Set the relay chain using a nested [`RelaychainConfigBuilder`].
     pub fn with_relaychain(
         self,
-        f: fn(
+        f: impl FnOnce(
             RelaychainConfigBuilder<relaychain::Initial>,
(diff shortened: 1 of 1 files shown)


pepoviola (approved):
Looks good, thanks!!

l0r1s:
Thanks @s0me0ne-unkn0wn !

PR #178, opened 2024-02-28
Explore integration test machinery

pepoviola:
Move integration tests to use `env var` to pick the provider.


pepoviola:
After talking with @l0r1s, we will be moving to impl `Drop` for Namespace to ensure resource cleanup.

pepoviola:
Closing in favor of https://github.com/paritytech/zombienet-sdk/pull/183

PR #180, opened 2024-03-03
Feat: impl db snap logic for k8s/native

pepoviola:
Add:
 - Logic to use db_snapshot in both native/k8s providers
 - Small fix on chain_spec_path (for both native/k8s)
 - New example `db_snapshot`

### DIFF
--- a/crates/orchestrator/src/spawner.rs
+++ b/crates/orchestrator/src/spawner.rs
@@ -156,7 +156,8 @@ where
                 .map(|var| (var.name.clone(), var.value.clone())),
         )
         .injected_files(files_to_inject)
-        .created_paths(created_paths);
+        .created_paths(created_paths)
+        .db_snapshot(node.db_snapshot.clone());
 
     let spawn_ops = if let Some(image) = node.image.as_ref() {
         spawn_ops.image(image.as_str())
--- a/crates/provider/src/kubernetes/node.rs
+++ b/crates/provider/src/kubernetes/node.rs
@@ -9,20 +9,21 @@ use std::{
 
 use anyhow::anyhow;
 use async_trait::async_trait;
-use configuration::shared::{constants::THIS_IS_A_BUG, resources::Resources};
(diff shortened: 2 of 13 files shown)


crates/orchestrator/src/spawner.rs
```diff
@@ -156,7 +156,8 @@ where
                 .map(|var| (var.name.clone(), var.value.clone())),
         )
         .injected_files(files_to_inject)
-        .created_paths(created_paths);
+        .created_paths(created_paths)
```
l0r1s, line 159: We can remove this, there is no need to use created_paths now that it's created on send_file

crates/provider/src/kubernetes/node.rs
```diff
@@ -293,6 +345,40 @@ where
...
+        })?;
+
+        if res.status() != reqwest::StatusCode::OK {
+            // we need to upload the file
+            self.http_client
+                .post(url.as_ref())
+                .body(data)
+                .send()
+                .await
+                .map_err(|err| {
```
l0r1s, line 374: What is the difference with ```send_file``` below ? You want to update without copying to the pod ? Then maybe split the logic below in 2 parts and have a function used where we need ? wdyt ?

l0r1s (approved):
👍

crates/orchestrator/src/spawner.rs
pepoviola: We still need this for `native`, we should abstract in some way for both providers in a follow up pr.

crates/provider/src/kubernetes/node.rs
pepoviola: Yes, sounds good. We should split the send_file (upload/download) since we don't always want to just copy to one pod.

crates/orchestrator/src/spawner.rs
l0r1s: You are right, we should probably at it here (as the Kubernetes provider): https://github.com/paritytech/zombienet-sdk/blob/2a59e27b36f09d1b5b838c6ad237fd662da55f14/crates/provider/src/native/node.rs#L408

PR #181, opened 2024-03-04
Feat/e2e tests native

l0r1s:
- Added a basic basic semi-blackbox test (we still need to know what ports/dir we use)
- Refacto a bit the node logs polling/writting by simplifyin and added error tracing
- Sorted deps in Cargo.toml (waiting for a merge in the crate https://github.com/DevinR528/cargo-sort to add it to the pipeline because it doesn't sort workspace.dependencies correctly atm)


pepoviola (commented):
Small comment, then looks good!!

PR #184, opened 2024-03-07
Feat/available args

l0r1s:
- Add logic to retrieve available args from node before  spawning the full network using the output of the command with --help

### DIFF
--- a/crates/orchestrator/src/lib.rs
+++ b/crates/orchestrator/src/lib.rs
@@ -85,6 +85,11 @@ where
         // create namespace
         let ns = self.provider.create_namespace().await?;
 
+        network_spec
+            .populate_nodes_available_args(ns.clone())
+            .await
+            .unwrap();
+
         info!("🧰 ns: {}", ns.name());
         info!("🧰 base_dir: {:?}", ns.base_dir());
 
--- a/crates/orchestrator/src/network_spec/node.rs
+++ b/crates/orchestrator/src/network_spec/node.rs
@@ -69,6 +69,9 @@ pub struct NodeSpec {
     /// Arguments to use for node. Appended to default.
     pub(crate) args: Vec<Arg>,
 
+    // The help command output containing the available arguments.
+    pub(crate) available_args_output: Option<String>,
+
     /// Wether the node is a validator.
     pub(crate) is_validator: bool,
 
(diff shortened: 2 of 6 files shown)


crates/orchestrator/src/lib.rs
```diff
@@ -85,6 +85,11 @@ where
         // create namespace
         let ns = self.provider.create_namespace().await?;
 
+        network_spec
+            .populate_nodes_available_args(ns.clone())
+            .await
+            .unwrap();
```
pepoviola, line 91: We should handle the error here instead of `unwrap`

pepoviola, line 91:
```suggestion
            .await?;
```

crates/orchestrator/src/network_spec/node.rs
```diff
@@ -260,4 +265,11 @@ impl NodeSpec {
             p2p_port: generators::generate_node_port(options.p2p_port)?,
         })
     }
+
+    pub(crate) fn supports_arg(&self, arg: impl AsRef<str>) -> bool {
+        self.available_args_output
+            .as_ref()
+            .expect("available args should be present at this point")
```
pepoviola, line 272: We should include `THIS_IS_A_BUG` here

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -363,6 +363,32 @@ where
             .collect()
     }
 
+    async fn get_node_available_args(
+        &self,
+        (command, image): (String, Option<String>),
+    ) -> Result<String, ProviderError> {
+        let node_image = image.expect(
+            "image should be present when getting node available args with kubernetes provider",
```
pepoviola, line 371: Also here

pepoviola (commented):
Looks good! some small comments and is good to merge :)

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         uses: taiki-e/install-action@cargo-llvm-cov
 
       - name: Collect coverage data
-        run: cargo llvm-cov nextest --lcov --output-path lcov.info
+        run: cargo llvm-cov nextest --workspace --exclude zombienet-sdk --test-threads 1 --lcov --output-path lcov.info
```
pepoviola, line 86: nice!

pepoviola (approved):

PR #186, opened 2024-03-08
Fix argument passing

s0me0ne-unkn0wn:
Concatenating the command line option and its value doesn't really work. The shell treats `--opt val` as a single argument `--opt val` instead of `--opt` `val` and fails to execute. This PR fixes that by separating options from their values.

Error example:
```
error: unexpected argument '--workers-path ~/.cargo/bin/workers-1.6' found

  tip: a similar argument exists: '--workers-path'
```


pepoviola (approved):

pepoviola:
Looks good @s0me0ne-unkn0wn, thanks!!

PR #189, opened 2024-03-16
Generalize metric assertions

s0me0ne-unkn0wn:
Enables using an arbitrary predicate to assert on metrics instead of just equality, which may be brittle because of casting everything to a floating-point type.

Usage example:

```rust
if node.assert_with("block_height{status=\"best\"}", |bh| bh > 10).await? {
    println!("Block height is over 10");
}
```


pepoviola (approved):
Looks great, thanks!!

PR #198, opened 2024-03-27
Feat/docker provider

l0r1s:
WIP


pepoviola:
Hi @l0r1s, looks great. What's the current status. I can continue with the implementation.

Thx!!

crates/provider/src/docker/client.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        if !result.status.success() {
+            return Err(anyhow!(
+                "Failed to detect container engine: {}",
+                String::from_utf8_lossy(&result.stderr)
+            )
+            .into());
+        }
+
+        Ok(String::from_utf8_lossy(&result.stdout).contains("podman"))
```
pepoviola, line 180: This will require that the user have at least a symlink `docker -> podman`, why not:
- Allow users to pass the `main cmd` (docker/podman) in `fn new`
- or, detect one or the other. Trying first `docker` and then `podman`

Thx!

crates/provider/src/docker/client.rs
```diff
@@ -0,0 +1,444 @@
...
+                    }
+
+                    None
+                },
+            })
+            .collect();
+
+        for name in container_names {
+            self.container_rm(&name).await?;
+        }
```
pepoviola, line 387: We can remove all containers at ones

```
docker rm --help

Usage:  docker rm [OPTIONS] CONTAINER [CONTAINER...]

Remove one or more containers
```

crates/provider/src/docker/client.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        let containers = serde_json::from_str(&stdout)
+            .map_err(|err| anyhow!("Failed to parse podman containers output: {err}"))?;
+
+        Ok(containers)
+    }
+
+    async fn get_docker_containers(&self) -> Result<Vec<DockerContainer>> {
+        let res = tokio::process::Command::new("docker")
+            .args(vec!["ps", "--all", "--no-trunc", "--format", "json"])
```
pepoviola, line 427: We should add some filter to get only the containers of the `ns`
`--filter name=<ns>` will work

crates/provider/src/docker/client.rs
```diff
@@ -0,0 +1,444 @@
...
+                .map(Container::Docker)
+                .collect()
+        };
+
+        Ok(containers)
+    }
+
+    async fn get_podman_containers(&self) -> Result<Vec<PodmanContainer>> {
+        let res = tokio::process::Command::new("docker")
+            .args(vec!["ps", "--all", "--no-trunc", "--format", "json"])
```
pepoviola, line 412: Same here, we should filter the containers.

crates/provider/src/docker/namespace.rs
```diff
@@ -0,0 +1,356 @@
...
+            .create_volume(&zombie_wrapper_volume_name)
+            .await
+            .map_err(|err| ProviderError::CreateNamespaceFailed(self.name.clone(), err.into()))?;
+
+        // copy script to volume
+        self.docker_client
+            .container_run(
+                ContainerRunOptions::new(
+                    "alpine:latest",
+                    vec!["cp", "/zombie-wrapper.sh", "/scripts"],
```
pepoviola, line 109: Can we just add `"&&", "chmod", "-R", "777", "/scripts"` to _not_ need to create a new container for just this op.

crates/provider/src/docker/namespace.rs
```diff
@@ -0,0 +1,356 @@
...
+            .create_volume(&helper_binaries_volume_name)
+            .await
+            .map_err(|err| ProviderError::CreateNamespaceFailed(self.name.clone(), err.into()))?;
+
+        // download binaries to volume
+        self.docker_client
+            .container_run(
+                ContainerRunOptions::new(
+                    "alpine:latest",
+                    vec!["ash", "/helper-binaries-downloader.sh"],
```
pepoviola, line 164: Same here.

crates/provider/src/docker/node.rs
```diff
@@ -0,0 +1,527 @@
...
+        let command = [vec![self.program.to_string()], self.args.to_vec()].concat();
+
+        self.docker_client
+            .container_run(
+                ContainerRunOptions::new(&self.image, command)
+                    .name(&self.container_name)
+                    .env(self.env.clone())
+                    .volume_mounts(HashMap::from([
+                        (
+                            format!("{}-zombie-wrapper", self.namespace_name(),),
```
pepoviola, line 139:
```suggestion
                            format!("{}-zombie-wrapper", self.namespace_name()),
```

crates/provider/src/docker/node.rs
```diff
@@ -0,0 +1,527 @@
...
+        .await?;
+
+        Ok(())
+    }
+
+    pub(super) async fn start(&self) -> Result<(), ProviderError> {
+        self.docker_client
+            .container_exec(
+                &self.container_name,
+                vec!["sh", "-c", "echo", ">", "/tmp/zombiepipe"],
```
pepoviola, line 227:
```suggestion
                vec!["sh", "-c", "echo start > /tmp/zombiepipe"],
```

crates/provider/src/docker/node.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        Ok(())
+    }
+
+    async fn receive_file(
+        &self,
+        _remote_src: &Path,
+        _local_dest: &Path,
+    ) -> Result<(), ProviderError> {
+        Ok(())
```
pepoviola, line 447:
```suggestion
          todo!()
//        Ok(())
```

crates/provider/src/docker/node.rs
pepoviola: In other methods we are just using `["echo", "<method>", ">",  "/tmp/zombiepipe"] we should change here also.

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -385,11 +385,13 @@ where
...
             })?;
 
+        temp_node.destroy().await?;
+
         Ok(available_args_output)
     }
 
     async fn spawn_node(&self, options: &SpawnNodeOptions) -> Result<DynNode, ProviderError> {
-        trace!("spawn option {:?}", options);
+        trace!("spawn node options {options:?}");
```
pepoviola, line 394: 👍

crates/provider/src/kubernetes/namespace.rs
```diff
@@ -419,7 +421,7 @@ where
     }
 
     async fn generate_files(&self, options: GenerateFilesOptions) -> Result<(), ProviderError> {
-        debug!("options {:#?}", options);
+        debug!("generate files options {options:#?}");
```
pepoviola, line 424: We should remove this `pretty` debugs, since the multiline output isn't optimal.

crates/provider/src/shared/scripts/helper-binaries-downloader.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/ash
+
+log() {
+    echo "$(date +"%F %T") $1"
+}
+
+# used to handle the distinction where /cfg is used for k8s and /helpers for docker/podman
+# to share a volume across nodes containing helper binaries and independent from /cfg
+# where some node files are stored
+OUTDIR=$([ -d /helpers ] && echo "/helpers" || echo "/cfg")
```
pepoviola, line 10: This should be _noticed_ when we run scripts since we relay on the default `/cfg/[curl/coreutils]`

PR #204, opened 2024-04-27
feat: add support to build chain-spec with a custom command

pepoviola:


crates/configuration/testing/snapshots/0000-small-network.toml
```diff
@@ -14,6 +14,7 @@ validator = true
 invulnerable = true
 bootnode = false
 balance = 2000000000000
+args = [""]
```
pepoviola, line 17:
```suggestion
```

PR #209, opened 2024-05-13
feat: enhance public api

evilrobot-01:
Firstly, thank you for an amazing library! The original zombienet is a Polkadot staple and now being able to use it completely from Rust is a game changer for us: as you may be aware, we have integrated zombienet-sdk with [pop-cli](https://github.com/paritytech/zombienet-sdk).

In doing so, we needed to add a few getters to the public api to be able to present certain information to the user, as well we expose a little more of the inner errors. This PR simply upstreams these changes.

It also includes two additional fixes, which may be better implemented based on your guidance:
- collator args werent working for us when specified in a config file, which seemed to be due to missing `--` in the args list provided by calling functions. 
- artifact path wasnt set when trying to build a chain spec. This may no longer be required based on the changes in https://github.com/paritytech/zombienet-sdk/commit/c7ce50c0e54c7efd505d18c0d77fe169f7df0465

I am happy to add tests if you are open to these changes. 🙂

### DIFF
--- a/crates/orchestrator/src/errors.rs
+++ b/crates/orchestrator/src/errors.rs
@@ -18,7 +18,7 @@ pub enum OrchestratorError {
     InvariantError(&'static str),
     #[error("Global network spawn timeout: {0} secs")]
     GlobalTimeOut(u32),
-    #[error("Generator error")]
+    #[error("Generator error: {0}")]
     GeneratorError(#[from] generators::errors::GeneratorError),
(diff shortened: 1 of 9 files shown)


pepoviola:
Hi @evilrobot-01, thanks for your feedback, the fixes and for using zombienet-sdk 🙌 . Changes in the api looks good to me, I will merge those and publish a new version. 

Thanks!!

PR #211, opened 2024-05-14
fix(docker): don't early exit if docker binary isn't available

andresilva:


crates/provider/src/docker/client.rs
```diff
@@ -186,21 +186,24 @@ impl DockerClient {
     }
 
     async fn is_using_podman() -> Result<bool> {
-        let result = tokio::process::Command::new("docker")
+        if let Ok(output) = tokio::process::Command::new("docker")
+            .arg("version")
```
andresilva, line 190: Using docker alias `docker --version` wasn't returning any string containing podman. Using `docker version` I get an output like this:

```
❯ docker version 
Client:       Podman Engine
Version:      5.0.2
API Version:  5.0.2
Go Version:   go1.22.2
Built:        Tue Jan  1 00:00:00 1980
OS/Arch:      linux/amd64
```

andresilva:
I tested this locally and it worked with podman (and also using docker symlink to podman). I didn't test with proper docker though.

pepoviola:
Looks good to me! thanks for the fix, I will testing in `docker` before  merge it.

Thanks!!

PR #212, opened 2024-05-14
fix(docker): don't use docker command if using podman

andresilva:


andresilva:
With this PR I was able to spawn a network without any docker symlink (i.e. just using podman directly).

pepoviola (approved):
Looks great, thanks!!

PR #216, opened 2024-05-16
feat: put `pjs-rs` behind a feature-gate

pepoviola:
cc @sandreim thanks for the suggestion :)


evilrobot-01:
Great to see this, thanks!

PR #243, opened 2024-07-30
reuse pjs code execution code

ozgunozerk:
This is nothing big  :) 

The code execution logic is already present under `pjs` function, yet it is duplicated for `pjs_file` function. 

This change re-uses the existing code, so that you won't have to change it twice in the future.


pepoviola (approved):
thanks!!

PR #244, opened 2024-08-12
fix(chain-spec): adjust hrmp channels json structure

evilrobot-01:
When testing `zombienet-sdk` with cross-chain calls with Asset Hub, we found that the hrmp channels passed to the chainspec caused the following issue:

`Error: "Invalid JSON blob: invalid type: map, expected a tuple of size 4 at line 1 column 2794"`

This is because the values of `HrmpChannelConfig` serialise to a map instead of a tuple, as expected by https://github.com/paritytech/polkadot-sdk/blob/fc906d5d0fb7796ef54ba670101cf37b0aad6794/polkadot/runtime/parachains/src/hrmp.rs#L492. Tests were updated accordingly to facilitate this fix.

Additional commits were also added to enable some of Asset Hub on Polkadot functionality observed in the classic Zombienet repo, so that we are able to launch a local Polkadot network with Asset Hub should anyone require. Example configuration files can be found at the bottom of https://github.com/r0gue-io/pop-cli/pull/278/files for anyone interested, although that PR is still a work in progress and dependent on this one being accepted.

PS - I added tests where I could, but additional tests might require some minor refactoring to make it easier to test only the logic that is changed by this PR, which I am not sure what the appetite is? An example would be refactoring the file name generation in `keystore.rs` into its own function to more easily test the outputs without the `scoped_fs`.

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -921,9 +921,11 @@ fn add_authorities(
     nodes: &[&NodeSpec],
     use_stash: bool,
 ) {
-    let asset_hub_polkadot = chain_spec_json.get("id")
+    let asset_hub_polkadot = chain_spec_json
+        .get("id")
         .and_then(|v| v.as_str())
-        .map(|id| id.starts_with("asset-hub-polkadot")).unwrap_or_default();
+        .map(|id| id.starts_with("asset-hub-polkadot"))
+        .unwrap_or_default();
     if let Some(val) = chain_spec_json.pointer_mut(runtime_config_ptr) {
         let keys: Vec<GenesisNodeKey> = nodes
             .iter()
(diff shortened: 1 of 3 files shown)


pepoviola (approved):
Looks good! Thanks for fixing this. And also, yes. We need to make some refactors to make the code more modular and easy to test.
Thx!!

pepoviola:
Hi @evilrobot-01, can you format the code with
```
cargo +nightly fmt  --all
```
Thx!

evilrobot-01:
Yes, will do.

Apologies, I didn't want to mess with the formatting at the time and forgot to come back to it!

pepoviola:
> Yes, will do.
> 
> Apologies, I didn't want to mess with the formatting at the time and forgot to come back to it!

No problem 👍, thanks for your contribution and help :)

pepoviola:
Merged, thanks @evilrobot-01 🙌

Issue #251, opened 2024-08-28
Simplify Zombienet SDK DX


ozgunozerk:
### Is your feature request related to a problem? Please describe.

Current state of `zombienet-sdk` is feature-rich, but lacks ease of use for many setting that can be hidden for most of the users by utilizing defaults.

Simplifying the devx may increase and improve adoption, popularity, and open-source contribution.

### Describe the solution you'd like

What I have in mind is roughly:

```rust
let alice = ZombieNode::new("alice");
let bob = ZombieNode::new("bob");

let collator1 = ZombieCollator::new("collator1");
let collator2 = ZombieCollator::new("collator2");

let parachain = ZombieParachain::new().set_id(100).set_collators(collator1, collator2);

let zombienet = Zombienet::new() // also possible to supply a `config` file here by calling `from()` method
		.set_nodes(alice, bob)
		.set_parachains(parachain)
		.spawn();
	
let alice_client: subxt_client = zombienet.get_client("alice");

/*
	Here, we can use `alice_client` to submit any transaction or query the blockchain,
	any subxt (polkadotjs) functionality is exposed via `alice_client` here
*/
```

### Additional context

We met with @pepoviola and we are on the same page. 

We want to preserve the `builder` and `type-state-pattern`, to enforce compile time checks and errors for the users. 

This feature will not rewrite everything from scratch, but will act as a wrapper on top of the current code. Thus, preserve the safety measures that are already there

ozgunozerk:
Feel free to assign this issue to me. 

I'll submit a draft PR to have the discussion there and keep this issue compact for newcomers.

jmg-duarte:
On this, I'd re-export all `zombienet_*` crates with the SDK, instead of depending on them but requiring them to be explicitly added to the project instead.

I keep needing to add more stuff to the `Cargo.toml` that *should just be in the SDK*

ozgunozerk:
I don't have any strong preferences, as you guys see fit! @pepoviola I want to start on the implementation, did you have a chance to check out the draft PR for the more elaborate and concrete API design?

ozgunozerk:
Update: I have something ready, will polish it and submit it soon!

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/milestone-3-status-update-milestone-4-proposal-openzeppelin-x-polkadot-ecosystem-growth/10300/1

PR #252, opened 2024-08-28
Zombienet Simplified DevX

ozgunozerk:
Addresses #251 

This is the first stage of the changes. Provides the design of what can be done in order to simplify the developer experience.

Heavily utilizes `type-state-pattern` and somewhat `builder` pattern.

Due to `type-state-pattern`, some function body's are duplicated. The only way to make them re-usable is to utilize macros. The decision on whether to use macro's for that is completely left to @pepoviola as I don't have any strong preference.

@pepoviola, please share your comments if you want to make any changes to this design. After we both agree on the design, I can start writing the bodies of the functions.

Thanks in advance for being open to this improvement!


crates/wrapper/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn otherMethod(mut self) -> Self {}
+}
+
+
+use std::marker::PhantomData;
+
+struct EmptyState;
+struct IdOnly;
+struct CollatorOnly;
+struct IdAndCollator;
```
jmg-duarte, line 45: Downstream users can extend state sets and possibly break their guarantees. If you're looking to stop that, take a look into the following:

* https://rust-lang.github.io/api-guidelines/future-proofing.html
* https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/
* https://docs.rs/sealed/latest/sealed/

jmg-duarte (commented):
If I may add my two cents:

> Due to type-state-pattern, some function body's are duplicated. The only way to make them re-usable is to utilize macros. The decision on whether to use macro's for that is completely left to @pepoviola as I don't have any strong preference.

While the typestate pattern makes the builder safe by construction, I believe it also makes the DX worse for both developers and users and as such it should be kept only for very critical parts of the system.

For the builder system, I think validating everything in the end is enough, after all, the current builder still returns a `Result`, indicating that even the typestate pattern fails to fully achieve it's goal
```
    /// Seals the builder and returns a [`NetworkConfig`] if there are no validation errors, else returns errors.
    pub fn build(self) -> Result<NetworkConfig, Vec<anyhow::Error>> {
        if !self.errors.is_empty() {
            return Err(self.errors);
        }

        Ok(self.config)
    }
```

crates/wrapper/src/lib.rs
ozgunozerk: I know about sealed traits and was looking for a good place to utilize them, this will be fun!

> While the typestate pattern makes the builder safe by construction, I believe it also makes the DX worse for both developers and users and as such it should be kept only for very critical parts of the system.

Completely agreed. In our convo with @pepoviola we were praising type-state pattern a lot, and maybe I overdid a bit :D 

I'll tune it down for the DX and try to utilize sealed traits instead. 

These are fantastic suggestions, appreciated 👍

---

I think we all are more or less on the same page here, so I'll start the implementation after revizing this

jmg-duarte:
@ozgunozerk sounds great! Can't wait to see the result!

pepoviola:
Hi @ozgunozerk just to clarify, you a thinking in remove the type-state pattern from the `configuration` create and replace the builders with simple ones? Or you think in wrapping those from the `sdk`  crate?
I think simplify the external interface with simplest builders is the best path. Looking forward to see the impl

Thx!!

ozgunozerk:
@pepoviola yes, I'm thinking about using the simplest way for the wrapper, not planning to change anything internally

pepoviola:
> @pepoviola yes, I'm thinking about using the simplest way for the wrapper, not planning to change anything internally

Awesome 🙌, thanks!

ozgunozerk:
Closing due to #261

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/milestone-3-status-update-milestone-4-proposal-openzeppelin-x-polkadot-ecosystem-growth/10300/1

Issue #247, opened 2024-08-19
Allow to use `evm` based chains


pepoviola:
We should support `evm` based chains as we do in `v1` (see: https://github.com/paritytech/zombienet/issues/1826#issuecomment-2293690707). 

We can port the logic from there and use a config flag like `evm-based` or also infer the key in the chain-spec from the default one generated.

cc: @AlexD10S / @al3mart

PR #256, opened 2024-09-11
add support for evm and fix stash account

pepoviola:
- Fix stash derivation (use `//`)
- Add `evm` support, you can set your parachain as evm based with `evm_based(choice: bool)`.

Fix: #247 
cc: @AlexD10S / @al3mart


AlexD10S:
That works great for me! Thanks for adding this feature.

PR #266, opened 2024-10-03
add validation to node name, can't be empty

pepoviola:
Add: 
  - Validate `node_name`, can't be empty

cc: @ozgunozerk

### DIFF
--- a/crates/configuration/src/shared/helpers.rs
+++ b/crates/configuration/src/shared/helpers.rs
@@ -44,9 +44,7 @@ pub fn ensure_node_name_unique(
     Err(ValidationError::NodeNameAlreadyUsed(node_name).into())
 }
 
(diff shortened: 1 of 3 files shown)


crates/configuration/src/shared/helpers.rs
```diff
@@ -28,13 +28,14 @@ pub fn merge_errors_vecs(
 }
 
 pub fn ensure_node_name_unique(
-    node_name: String,
+    node_name: impl Into<String>,
```
ozgunozerk, line 31: `AsRef<str>` might be an even better candidate across the crate for this, accepting both `&str` and `String`, but this is a nitpick

pepoviola: Ahh yes, I will change in a follow pr. Thanks!

PR #270, opened 2024-10-25
refactor exports from `sdk`

pepoviola:
Refactor some exports from `sdk` in order to build easily on top of the `sdk` directly.

From the `sdk` we expose now

```
/// Provider types to construct and work with it
provider::

/// Shared type from inner-crates
shared::
  configuration::
  provider::
  support::
```

cc: @ozgunozerk / @jmg-duarte

### DIFF
--- a/crates/file-server/src/main.rs
+++ b/crates/file-server/src/main.rs
@@ -45,7 +45,8 @@ async fn main() {
         .await
         .expect(&format!("failed to listen on {address}"));
     tracing::info!("file server started on {}", listener.local_addr().unwrap());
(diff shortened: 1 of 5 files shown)


jmg-duarte:
I don't think you should prepend them with `shared`. They're part of the functionality of the sdk after all.

`zombienet_sdk::shared::orchestrator` seems weird while `zombienet_sdk::orchestrator` makes complete sense to me.

This also makes it easier for you since you can simply `pub use provider` and trust its privacy levels to solve the rest.

PR #275, opened 2024-11-25
Override keys iff they are present

pepoviola:
fix for https://github.com/paritytech/polkadot-sdk/issues/6575

cc: @iulianbarbu


iulianbarbu (approved):
LGTM!

PR #290, opened 2025-01-22
Deprecate `client` function and replace with `try_client`

bkchr:
Generally the idea is that `wait_client` should be used as a node can take a moment to appear online.


pepoviola (approved):
Looks great, thanks!! I will fix the clippy errors and merge.
Thx!

PR #291, opened 2025-01-29
fix: override genesis to work with toml files/api

pepoviola:
Add logic to allow overrides working with network definition in `toml` format and sdk api.

cc: @iulianbarbu

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -830,7 +830,7 @@ fn percolate_overrides<'a>(
     };
 
     let p = if i == pointer_parts.len() - 1 {
-        // top level key is at last part of the pointer
+        // top level key is at end of the pointer
         let p = format!("/{}", pointer_parts[i]);
         trace!("overrides pointer {p}");
         p
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -442,8 +442,10 @@ impl ChainSpec {
 
             // make genesis overrides first.
             if let Some(overrides) = &para.genesis_overrides {
+                let percolated_overrides = percolate_overrides(&pointer, overrides)
+                    .map_err(|e| GeneratorError::ChainSpecGeneration(e.to_string()))?;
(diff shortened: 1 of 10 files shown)


crates/orchestrator/src/generators/chain_spec.rs
```diff
@@ -803,6 +807,77 @@ fn get_runtime_config_pointer(chain_spec_json: &serde_json::Value) -> Result<Str
...
+    trace!("top_level_key: {top_level_key}");
+    let index = pointer_parts.iter().position(|x| *x == top_level_key);
+    let Some(i) = index else {
+        return Err(anyhow!(
+            "Top level key {top_level_key} should be in the pointer: {pointer}"
+        ));
+    };
+
+    let p = if i == pointer_parts.len() - 1 {
+        // top level key is at last part of the pointer
```
iulianbarbu, line 833: nit:
```suggestion
        // top level key is at end of the pointer
```

crates/orchestrator/src/generators/chain_spec.rs
```diff
@@ -1158,6 +1243,42 @@ mod tests {
...
+        use serde::{Deserialize, Serialize};
+
+        #[derive(Debug, Serialize, Deserialize)]
+        struct MockConfig {
+            #[serde(rename = "genesis", skip_serializing_if = "Option::is_none")]
+            genesis_overrides: Option<serde_json::Value>,
+        }
+
+        let mut chain_spec_json = chain_spec_test(ROCOCO_LOCAL_PLAIN_TESTING);
+        // Could also be  something like [genesis.runtimeGenesis.patch.balances]
```
iulianbarbu, line 1257: nit:
Can we also add an example for this in the test, besides the dev accounts, and assert on its existence like we did with `dev accounts`?

iulianbarbu (approved):

crates/orchestrator/src/generators/chain_spec.rs
pepoviola: Yes, I have a todo for adding more tests/examples. But I think is ok to unlock you asap.

iulianbarbu: Sounds good! thx!

PR #320, opened 2025-03-28
[native] fix destroy network on native provider

pepoviola:


crates/provider/src/native/namespace.rs
```diff
@@ -214,8 +214,18 @@ where
     }
 
     async fn destroy(&self) -> Result<(), ProviderError> {
+
+
```
pepoviola, line 218:
```suggestion
```

pepoviola, line 218:
```suggestion
```

PR #333, opened 2025-05-02
Log lines parsing improvements

lrubasze:
This PR improves log parsing by introducing some changes `wait_log_line_count_with_timeout()`.

The method searches log lines for a given substring or glob pattern,
and evaluates the number of matching lines using a user-provided predicate function.
Optionally, it can wait for the full timeout duration to ensure the condition
holds consistently (e.g., for verifying absence of logs).

It uses `LogLineCountOptions` struct to configure the timeout, match count predicate and full-duration waiting.

It returns enum `LogLineCount`:
* `Ok(LogLineCount::TargetReached(n))` if the predicate was satisfied within the timeout,
* `Ok(LogLineCount::TargetFails(n))` if the predicate was not satisfied in time,
* `Err(e)` if an error occurred during log retrieval or matching.

Additionally added some tests to verify the changes.

NOTE!
This is a breaking change (method signature changed) and the clients using `wait_log_line_count_with_timeout()` update their tests.

### DIFF
--- a/crates/orchestrator/src/network/node.rs
+++ b/crates/orchestrator/src/network/node.rs
@@ -370,6 +370,10 @@ impl NetworkNode {
 
     /// Wait until a the number of matching log lines is reach
     /// with timeout (secs)
+    #[deprecated(
+        since = "0.2.32",
+        note = "Use `wait_log_line_count_with_timeout_v2` instead"
+    )]
     pub async fn wait_log_line_count_with_timeout(
         &self,
         substring: impl Into<String>,
--- a/crates/orchestrator/src/network/node.rs
+++ b/crates/orchestrator/src/network/node.rs
@@ -37,6 +37,12 @@ pub struct NetworkNode {
     metrics_cache: Arc<RwLock<MetricMap>>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LogLineCount {
+    TargetReached(u32, u64), // log lines count, duration
+    TargetFailed(u32),       // log lines count
+}
+
 // #[derive(Clone, Debug)]
 // pub struct QueryMetricOptions {
 //     use_cache: bool,
@@ -355,6 +361,65 @@ impl NetworkNode {
         .await?
     }
 
+    /// Wait until a the number of matching log lines is reach
+    /// with timeout (secs)
+    pub async fn wait_log_line_count_with_timeout_v2(
+        &self,
+        substring: impl Into<String>,
+        is_glob: bool,
+        predicate: impl Fn(u32) -> bool,
+        timeout_secs: impl Into<u64>,
+        wait_until_timeout_elapses: bool,
+    ) -> Result<LogLineCount, anyhow::Error> {
+        let secs = timeout_secs.into();
+        let substring = substring.into();
+        debug!("waiting until match lines count within {secs} seconds");
+
+        let start = tokio::time::Instant::now();
+
+        let match_fn: Box<dyn Fn(&str) -> bool + Send + Sync> = if is_glob {
+            Box::new(move |line: &str| glob_match(&substring, line))
+        } else {
+            let re = Regex::new(&substring)?;
+            Box::new(move |line: &str| re.is_match(line))
+        };
+
+        let deadline = Duration::from_secs(secs);
+
+        if wait_until_timeout_elapses {
+            tokio::time::sleep(Duration::from_secs(secs)).await;
+        }
+
+        let mut q;
+        loop {
+            q = 0_u32;
+            let logs = self.logs().await?;
+            println!("logs = {logs:?}");
+            for line in logs.lines() {
+                if match_fn(line) {
+                    q += 1;
+                    if !wait_until_timeout_elapses {
+                        if predicate(q) {
+                            return Ok(LogLineCount::TargetReached(q, start.elapsed().as_secs()));
+                        }
+                    }
+                }
+            }
+
+            if start.elapsed() >= deadline {
+                break;
+            }
+
+            tokio::time::sleep(Duration::from_secs(2)).await;
+        }
+
+        if predicate(q) {
+            return Ok(LogLineCount::TargetReached(q, start.elapsed().as_secs()));
+        } else {
+            return Ok(LogLineCount::TargetFailed(q));
+        }
+    }
+
     // TODO: impl
     // wait_event_count
     // wait_event_count_with_timeout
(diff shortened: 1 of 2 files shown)


crates/orchestrator/src/network/node.rs
```diff
@@ -37,6 +37,59 @@ pub struct NetworkNode {
...
+///
+/// # Variants
+/// - `TargetReached(count, duration_ms)` – The predicate condition was satisfied within the timeout.
+///     * `count`: The number of matching log lines at the time of satisfaction.
+///     * `duration_ms`: Time elapsed (in milliseconds) until the condition was met.
+/// - `TargetFailed(count)` – The condition was not met within the timeout.
+///     * `count`: The final number of matching log lines at timeout expiration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LogLineCount {
+    TargetReached(u32, u64),
```
pepoviola, line 52: out of curiosity,  did you have a use-case that needs the amount of time taken to reach the target?

crates/orchestrator/src/network/node.rs
```diff
@@ -339,6 +392,10 @@ impl NetworkNode {
 
     /// Wait until a the number of matching log lines is reach
     /// with timeout (secs)
+    #[deprecated(
+        since = "0.2.32",
+        note = "Use `wait_log_line_count_with_timeout_v2` instead"
```
pepoviola, line 397: nice 🙌

crates/orchestrator/src/network/node.rs
```diff
@@ -355,6 +412,101 @@ impl NetworkNode {
...
+        }
+
+        let mut q;
+        loop {
+            q = 0_u32;
+            let logs = self.logs().await?;
+            for line in logs.lines() {
+                if match_fn(line) {
+                    q += 1;
+                    if !options.wait_until_timeout_elapses && (options.predicate)(q) {
```
pepoviola, line 490: if `options. wait_until_timeout_elapses` is true we already _waited_ the desired time so we can just use 

```suggestion
                    (options.predicate)(q) {
```

crates/orchestrator/src/network/node.rs
```diff
@@ -441,4 +593,376 @@ impl std::fmt::Debug for NetworkNode {
     }
 }
 
-// TODO: mock and impl unit tests
+// TODO: mock and impl more unit tests
+#[cfg(test)]
```
pepoviola, line 597: Awesome! thanks for also cover this old _TODO_

crates/orchestrator/src/network/node.rs
lrubasze: Actually no :) I will remove it.

crates/orchestrator/src/network/node.rs
lrubasze: That's true. But on the other hand if `options.wait_until_timeout_elapses` is true I thought we should check the condition after processing all available lines. Otherwise we could return `TargetReached` upon first time when predicate is true ignoring that in the next steps the predicate might become false.
Example:
- we search for exactly one line "system launched", we shall wait until timeout 2s elapses.
So we waited 2s before the loop and then in the loop we are processing the the log (eg. 1000 lines).
In line 100 we found the "system launched" and we immediately return `TargetReached` ignoring the rest of the log, which could include another "system launched" in line 900. So in total there are 2 lines matching the pattern but we expected exactly one.

I will add some comment to make it more clear.

crates/orchestrator/src/network/node.rs
lrubasze: Ha, just covered log parsing, but I think it might be a good starting point to add more tests :)

skunert:
Not a strong opinion, but do we expect that usage of the old method is widespread? zombienet-sdk is stil fairly fresh, maybe we can do a breaking change and skip the `v2`? Its a bit annoying but I would expect that the migration in polkadot-sdk is pretty easy.

Up to you guys however :)

lrubasze:
> Not a strong opinion, but do we expect that usage of the old method is widespread? zombienet-sdk is stil fairly fresh, maybe we can do a breaking change and skip the `v2`? Its a bit annoying but I would expect that the migration in polkadot-sdk is pretty easy.
> 
> Up to you guys however :)

Yes, migrating in `polkadot-sdk` is not a problem at all, I will take care of that. 
Though not sure about the other clients of `zombienet-sdk`.
 @pepoviola it's your call :)

crates/orchestrator/src/network/node.rs
pepoviola: ahh make sense, good catch!

pepoviola:
@skunert / @lrubasze, At the moment we don't have much external users for the `sdk` so we can just implement as breaking change and migrate the test in `polkadot-sdk` repo. @lrubasze can you make those changes?

Thanks!

crates/orchestrator/src/network/node.rs
```diff
@@ -337,22 +389,104 @@ impl NetworkNode {
...
-            self.wait_log_line_count(substring, is_glob, count),
-        )
-        .await?
+        options: LogLineCountOptions,
+    ) -> Result<LogLineCount, anyhow::Error> {
+        let secs = options.timeout_secs;
+        let substring = substring.into();
+        debug!("waiting until match lines count within {secs} seconds");
+
+        let start = tokio::time::Instant::now();
```
pepoviola, line 445: We can remove this since we are not returning anymore the _duration_
```suggestion

```

crates/orchestrator/src/network/node.rs
```diff
@@ -37,6 +37,58 @@ pub struct NetworkNode {
...
+/// - `predicate`: A function that takes the current number of matching lines and
+///   returns `true` if the condition is satisfied.
+/// - `timeout_secs`: Maximum number of seconds to wait.
+/// - `wait_until_timeout_elapses`: If `true`, the system will continue waiting
+///   for the full timeout duration, even if the condition is already met early.
+///   Useful when you need to verify sustained absence or stability (e.g., "ensure no new logs appear").
+#[derive(Clone)]
+pub struct LogLineCountOptions {
+    pub predicate: Arc<dyn Fn(u32) -> bool + Send + Sync>,
+    pub timeout_secs: u64,
```
skunert, line 70: nit: IMO this timeout should be a `Duration`. Makes the API a bit nicer.

skunert (approved):

crates/orchestrator/src/network/node.rs
```diff
@@ -37,6 +37,58 @@ pub struct NetworkNode {
...
+///
+/// Indicates whether the log line count condition was met within the timeout period.
+///
+/// # Variants
+/// - `TargetReached(count)` – The predicate condition was satisfied within the timeout.
+///     * `count`: The number of matching log lines at the time of satisfaction.
+/// - `TargetFailed(count)` – The condition was not met within the timeout.
+///     * `count`: The final number of matching log lines at timeout expiration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LogLineCount {
```
skunert, line 50: Random idea: WDYT about having a `target_reached` or `success` method. This would make it easier to to just assert success without needing to use `matches` like in your breaking changes PR.

lrubasze: Ha, actually I really like this idea. Will add it!

crates/orchestrator/src/network/node.rs
lrubasze: Will do :)

pepoviola: The `new` method accept `impl Into<u64>` but we can change the type to be a `Duration` for better _dx_

Thanks!

crates/orchestrator/src/network/node.rs
lrubasze: Done

PR #335, opened 2025-05-05
add logic to handle `devStakers` key without override the validator count

pepoviola:
Only clear `validatorCount` IFF staking is set but __not__ `devStakers`.

cc: @tdimitrov

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -1396,7 +1396,6 @@ mod tests {
         }})
     }
 
-
     fn chain_spec_with_dev_stakers() -> serde_json::Value {
         json!({"genesis": {
             "runtimeGenesis" : {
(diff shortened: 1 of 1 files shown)


crates/orchestrator/src/generators/chain_spec.rs
```diff
@@ -1393,6 +1396,37 @@ mod tests {
...
+                        ],
+                        "forceEra": "NotForcing",
+                        "invulnerables": [],
+                        "maxNominatorCount": null,
+                        "maxValidatorCount": null,
+                        "minNominatorBond": 0,
+                        "minValidatorBond": 0,
+                        "slashRewardFraction": 0,
+                        "stakers": [],
+                        "validatorCount": 500
```
tdimitrov, line 1423: Ideally we want to get this from the chainspec.

pepoviola: Yes, this is covered in here https://github.com/paritytech/zombienet-sdk/pull/335/files#diff-9cfbdbefcc94bdcfc60d2331da54800823b05adae8f62414ead42659497fa5f2R1449 :)

PR #341, opened 2025-05-14
Allow insecure urls for RPC clients

lrubasze:
This is especially useful for CI, which spins nodes using Kubernetes provider and nodes don't have localhost IP addresses, which are not considered secure. This was not the case for `native` provider, which uses localhost addresses.

### DIFF
--- a/crates/orchestrator/src/network/node.rs
+++ b/crates/orchestrator/src/network/node.rs
@@ -159,7 +159,11 @@ impl NetworkNode {
 
     /// Get the rpc client for the node
     pub async fn rpc(&self) -> Result<RpcClient, subxt::Error> {
-        RpcClient::from_url(&self.ws_uri).await
+        if subxt::utils::url_is_secure(&self.ws_uri)? {
+            RpcClient::from_url(&self.ws_uri).await
+        } else {
+            RpcClient::from_insecure_url(&self.ws_uri).await
+        }
     }
 
     /// Get the [online client](subxt::client::OnlineClient) for the node
--- a/crates/orchestrator/src/network/parachain.rs
+++ b/crates/orchestrator/src/network/parachain.rs
@@ -158,7 +158,12 @@ impl Parachain {
                     options.node_ws_url.as_str()
                 )
             })?;
-        let api = OnlineClient::<SubstrateConfig>::from_url(options.node_ws_url).await?;
+
+        let api = if subxt::utils::url_is_secure(&options.node_ws_url)? {
+            OnlineClient::<SubstrateConfig>::from_url(options.node_ws_url).await?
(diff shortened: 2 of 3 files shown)


crates/orchestrator/src/network/node.rs
```diff
@@ -159,7 +159,11 @@ impl NetworkNode {
 
     /// Get the rpc client for the node
     pub async fn rpc(&self) -> Result<RpcClient, subxt::Error> {
-        RpcClient::from_url(&self.ws_uri).await
+        if subxt::utils::url_is_secure(&self.ws_uri)? {
+            RpcClient::from_url(&self.ws_uri).await
+        } else {
+            RpcClient::from_insecure_url(&self.ws_uri).await
+        }
```
pepoviola, line 166: Nice catch, we use the same logic [here](https://github.com/paritytech/zombienet-sdk/blob/fd43dba31990aeb47fa30bcb1d4b20fb7e1afbbe/crates/orchestrator/src/network/node.rs#L188-L192), maybe we can abstract in some way and allow also to get a client/rpc_client by just passing the url.

crates/orchestrator/src/network/parachain.rs
```diff
@@ -158,7 +158,12 @@ impl Parachain {
                     options.node_ws_url.as_str()
                 )
             })?;
-        let api = OnlineClient::<SubstrateConfig>::from_url(options.node_ws_url).await?;
+
+        let api = if subxt::utils::url_is_secure(&options.node_ws_url)? {
+            OnlineClient::<SubstrateConfig>::from_url(options.node_ws_url).await?
+        } else {
+            OnlineClient::<SubstrateConfig>::from_insecure_url(options.node_ws_url).await?
+        };
```
pepoviola, line 166: I think there are two options here to make the code cleaner, 
1.  Receive a ref of the `node` as part of the options and call `api = node.wait_client()`
2. Create a helper that create a _client_ from the url, this could be in the helpers module.
Thx!

crates/orchestrator/src/network/node.rs
lrubasze: Added `ClientFromUrl` trait and `get_client_from_url()` helper method, which allows to get `RpcClient` or `OnlineClient`

crates/orchestrator/src/network/parachain.rs
lrubasze: Picked option 2 - using `get_client_from_url()` mentioned above.

pepoviola (approved):
Looks great, thanks!!

pepoviola:
@lrubasze published as ` v0.3.1` thanks!!

Issue #329, opened 2025-04-25
Make logs more accessible


skunert:
### Is your feature request related to a problem? Please describe.

The old zombienet binary was printing the URL to the logs in loki directly, this was super nice. These days I need to get the namespace from the log output and put it into loki manually :(.

### Describe the solution you'd like

- Print log location directly into the output as before
- Put the logs into the artifacts for direct downloads

pepoviola:
Hi @skunert, sorry for the delay. Just for the record, in the `sdk` we take the path of off-loading this _ci_ related features and build them in the _ci_ machinery. Currently we are putting the logs in the artifacts and we are working to send to loki.

Thanks!! 🙌

PR #343, opened 2025-05-15
Support dumping logs

lrubasze:
This PR adds support for dumping logs.

Logs are dumped into a respective folder (unified across providers) so they could be easily picked up eg. by CI job. 
Also if running in CI then Loki urls are generated and printed (`kubernetes` provider only).
For convenience logs are dumped upon Network drop.

This PR addresses https://github.com/paritytech/zombienet-sdk/issues/329


crates/orchestrator/src/network.rs
```diff
@@ -61,7 +62,27 @@ macros::create_add_options!(AddCollatorOptions {
...
+        });
+
+        if let Err(err) = result {
+            error!("Error shutting down the network - {:?}", err);
+        };
+    }
+}
+
 impl<T: FileSystem> Network<T> {
+    async fn shutdown(&self) -> Result<(), anyhow::Error> {
```
lrubasze, line 80: Not sure if this method is really needed.
Maybe I could just call `self.dump_logs()` 

```suggestion
     // Teardown the network
    pub async fn destroy(&self) -> Result<(), ProviderError> {
        self.dump_logs().await?;
        self.ns.destroy().await
    }```

File: crates/provider/src/kubernetes/node.rs [hunk 2]
```diff
@@ -538,15 +541,38 @@ where
...
         let logs = self.logs().await?;
 
         self.filesystem
-            .write(local_dest, logs)
+            .write(&local_dest, logs)
             .await
             .map_err(|err| ProviderError::DumpLogsFailed(self.name.to_string(), err.into()))?;
 
-        Ok(())
+        if running_in_ci() {
```
lrubasze, line 555: Assumed that we run CI only with `kubernetes`. Not sure if this is correct?

pepoviola:
Hi @lrubasze, thanks for your contribution. In the previous version we had a test runner that orchestrate all the assertion and  logs collection at the end, but we have several issues with k8s since errors dumping logs produce false/positives for example. That why with this new version we though that the machinery to dump the logs and print the links should not be attached to the _network_ (or at least not to the main test process).

My idea to provide that info is that we use the `zombie.json` file in to provide the links to grafana as other _step_ and we can also dump the logs as _other step_ too, by not shutting down the network in `CI` (since we automatically clean all the resources in the cluster).  __But__ this is on the context of our CI, I think for `podman` dumping the logs on drop could be handy since we are deleting the containers.

wdyt?

Thanks!!

lrubasze:
Closing since it was agreed to implement the way @pepoviola suggests

lrubasze:
As above

PR #346, opened 2025-05-19
Network support paras with the same id

lrubasze:
Support parachains with the same id in the spawned network.

@pepoviola

### DIFF
--- a/crates/sdk/tests/two-paras-same-id.rs
+++ b/crates/sdk/tests/two-paras-same-id.rs
@@ -1,0 +1,39 @@
+use zombienet_sdk::{NetworkConfigBuilder, NetworkConfigExt};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn two_paras_same_id() {
+    tracing_subscriber::fmt::init();
+    let network = NetworkConfigBuilder::new()
+        .with_relaychain(|r| {
+            r.with_chain("rococo-local")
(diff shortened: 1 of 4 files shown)


crates/sdk/tests/two-paras-same-id.rs
```diff
@@ -0,0 +1,44 @@
...
+        .with_parachain(|p| {
+            p.with_id(2000)
+                .with_default_command("polkadot-parachain")
+                .with_default_image("docker.io/parity/polkadot-parachain:1.7.0")
+                .with_registration_strategy(zombienet_sdk::RegistrationStrategy::Manual)
+                .with_collator(|n| n.with_name("collator1"))
+        })
+        .build()
+        .unwrap()
+        .spawn_native()
```
pepoviola, line 29: Let's set the provider using _env vars_ like we use [here](https://github.com/paritytech/zombienet-sdk/blob/main/crates/sdk/tests/smoke-native.rs#L24-L28)

pepoviola (commented):
Small comments about the test, the rest looks great!!

thanks!!

crates/sdk/tests/two-paras-same-id.rs
lrubasze: good point 👍 done!

pepoviola (approved):

PR #348, opened 2025-05-22
Improve context replacements

lrubasze:
# Changes:
- Switched to sequential node spawning in order to allow referring to other nodes within the same group
- Extend argument list with other argument list instead of adding arguments joined with " ", 

# More details
This is to cover cases  where nodes spawned in one shot one group refer to the other nodes within that group.
Example:
- `alice` - spawned separately as a bootnode, added to the context
- `bob`, `charlie`, `dave`
`dave` refers to `alice`, `bob` and `charlie`
  - **before** 
Spawned as remaining nodes in one shot. their info were added to the context after all of them were spawned.
   - **now**
Nodes spawned sequentially and the node info is added directly after node is spawned.

```rust
let config = NetworkConfigBuilder::new()
.with_relaychain(|r| {
	r.with_chain("rococo-local")
		.with_default_image(images.polkadot.as_str())
		.with_default_args(vec![("-lparachain=debug").into()])
		.with_node(|node| node.with_name("alice"))
		.with_node(|node| node.with_name("bob"))
		.with_node(|node| node.with_name("charlie"))
		.with_node(|node| {
			node.with_name("dave").with_args(vec![
				("-lparachain=debug").into(),
				("--no-beefy").into(),
				("--reserved-only").into(),
				(
					"--reserved-nodes",
					vec![
						"{{ZOMBIE:alice:multiaddr}}",
						"{{ZOMBIE:bob:multiaddr}}",
						"{{ZOMBIE:charlie:multiaddr}}",
					],
				).into(),
				("--sync", "warp").into(),
			])
		})
})
```

### DIFF
--- a/crates/orchestrator/src/generators/command.rs
+++ b/crates/orchestrator/src/generators/command.rs
@@ -73,9 +73,8 @@ pub fn generate_for_cumulus_node(
         let bootnodes = bootnodes_addresses
             .iter()
             .map(|m| m.to_string())
-            .collect::<Vec<String>>()
-            .join(" ");
-        tmp_args.push(bootnodes)
+            .collect::<Vec<String>>();
+        tmp_args.extend(bootnodes)
     }
 
     // ports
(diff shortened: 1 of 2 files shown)


pepoviola:
Hi @lrubasze, thanks for checking this. Since the cases where nodes refer each other in the config as this one is an small part in the past (v1) we take the approach to keep the spawning in batch by default but if we detect some _refs_ between them we switch to sequentially. So, we don't affect the performance of other to support this use-case. Also, we can add the _concurrency cli arg_ to match also the v1 cli.

wdyt?

lrubasze:
Sure, will refactor then :)

lrubasze:
Cont

PR #350, opened 2025-06-02
feat, allow spawn seq.

pepoviola:
Allow to set the concurrency for spawn nodes. _If_ we detect some reference between the nodes (using the `ZOMBIE:` token we override the concurrency to 1. 

Also, the concurrency can be set using this env var `ZOMBIE_SPAWN_CONCURRENCY` that take precedence over the config.

cc: @lrubasze

### DIFF
--- a/crates/configuration/src/global_settings.rs
+++ b/crates/configuration/src/global_settings.rs
@@ -39,6 +39,8 @@ pub struct GlobalSettings {
     /// Used to reuse the same files (database) from a previous run,
     /// also note that we will override the content of some of those files.
     base_dir: Option<PathBuf>,
+    /// Number of concurrent spawning process to launch, None means try to spawn all at the same time.
+    spawn_concurrency: Option<usize>,
 }
 
 impl GlobalSettings {
@@ -67,6 +69,11 @@ impl GlobalSettings {
     pub fn base_dir(&self) -> Option<&Path> {
         self.base_dir.as_deref()
     }
+
+    /// Number of concurrent spawning process to launch
+    pub fn spawn_concurrency(&self) -> Option<usize> {
+        self.spawn_concurrency
+    }
 }
 
 impl Default for GlobalSettings {
@@ -77,6 +84,7 @@ impl Default for GlobalSettings {
             node_spawn_timeout: default_node_spawn_timeout(),
             local_ip: Default::default(),
             base_dir: Default::default(),
+            spawn_concurrency: Default::default(),
         }
     }
 }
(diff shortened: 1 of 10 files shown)


crates/configuration/src/global_settings.rs
```diff
@@ -77,30 +84,26 @@ impl Default for GlobalSettings {
...
-                network_spawn_timeout: default_timeout(),
-                node_spawn_timeout: default_node_spawn_timeout(),
-                local_ip: None,
-                base_dir: None,
-            },
-            errors: vec![],
-        }
-    }
-}
+// impl Default for GlobalSettingsBuilder {
```
lrubasze, line 99: Nitpick, but probably worth removing

crates/orchestrator/src/lib.rs
```diff
@@ -733,4 +807,88 @@ mod tests {
...
+        node.args
+            .push("--bootnodes {{ZOMBIE:bob:multiAddress')}}".into());
+        let (concurrency, limited) = calculate_concurrency(&spec).unwrap();
+        assert_eq!(concurrency, 1);
+        assert!(limited);
+    }
+
+    #[tokio::test]
+    async fn set_spawn_concurrency_from_env() {
+        // let mutex = ENV_MUTEX;
```
lrubasze, line 866: Nitpick, but probably not needed

crates/orchestrator/src/lib.rs
```diff
@@ -733,4 +807,88 @@ mod tests {
...
+        let network_config = generate(false, Some("cargo")).unwrap();
+        let spec = NetworkSpec::from_config(&network_config).await.unwrap();
+        let (concurrency, limited) = calculate_concurrency(&spec).unwrap();
+        assert_eq!(concurrency, 10);
+        assert!(!limited);
+    }
+
+    #[tokio::test]
+    async fn set_spawn_concurrency_from_env_but_limited() {
+        // let mutex = ENV_MUTEX;
```
lrubasze, line 880: ditto

crates/orchestrator/src/lib.rs
```diff
@@ -637,10 +695,26 @@ pub use pjs_helper::PjsResult;
...
+    use configuration::{GlobalSettingsBuilder, NetworkConfigBuilder};
+    use lazy_static::lazy_static;
+    use tokio::sync::Mutex;
 
     use super::*;
 
+    const ENV_KEY: &str = "ZOMBIE_SPAWN_CONCURRENCY";
+    // mutex for test that use env
+    lazy_static! {
+        static ref ENV_MUTEX: Mutex<()> = Mutex::new(());
```
lrubasze, line 707: Is this mutex really needed?
Isn't it enough to make sure that env is set/removed before `calculate_frequency` is called?

lrubasze (commented):
Nice work! though one question popped up :)

crates/orchestrator/src/lib.rs
pepoviola: Yes, since the unit test doesn't run in order (and can be run in parallel) we need to ensure the proper env is set.

PR #359, opened 2025-06-04
`configuration::WithRelaychain` is now public

michalkucharczyk:
I wanted to define a following function which accepts partially configured `NetworkConfigBuilder` and sets sets the base dir.  I didn't manage to do this without re-exporting `configuration::WithRelaychain` type.

```rust
	pub async fn with_closure<F>(builder: F) -> Result<NetworkSpawner>
	where
		F: FnOnce() -> NetworkConfigBuilder<WithRelaychain>,
	{
		let _ = env_logger::try_init_from_env(
			env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "info"),
		);

		let config_builder = builder();

		let net_config = config_builder
			.with_global_settings(|global_settings| match NetworkSpawner::base_dir_from_env() {
				Some(val) => global_settings.with_base_dir(val),
				_ => global_settings,
			})
			.build()
			.unwrap();

		Ok(NetworkSpawner {
			network: net_config
				.spawn_native()
				.await
				.map_err(|err| Error::NetworkInit(anyhow!(err.to_string())))?,
		})
	}
```


michalkucharczyk:
Maybe there is no need to merge it, and there is other way to go.

michalkucharczyk:
A bit more context:
https://github.com/paritytech/polkadot-sdk/blob/dd91dfd3a24e4940abe8ac74d4abd8ed02305076/substrate/client/transaction-pool/tests/zombienet/mod.rs#L80-L108

And usage here:
https://github.com/paritytech/polkadot-sdk/blob/dd91dfd3a24e4940abe8ac74d4abd8ed02305076/substrate/client/transaction-pool/tests/zombienet/yap_test.rs#L37

pepoviola:
Hey @michalkucharczyk, thanks for your feedback 🙌 . I think this is a bug [here](https://github.com/paritytech/zombienet-sdk/blob/main/crates/configuration/src/network.rs#L398). The global_settings should be allowed to set at any _state_ of the builder. Let me fix it and incorporate it in your branch. wdyt?

Thx!

michalkucharczyk:
@pepoviola please just relase the fix, and I will handle it by myself in my branch. Just ping me here. Thanks!

pepoviola (approved):

pepoviola:
@michalkucharczyk published as `v0.3.5`. Thanks 🙌 !

Issue #367, opened 2025-06-30
The same port is assigned to two different nodes


lrubasze:
### Issue Description

Issue occurred when running `zombienet-cumulus-0002-pov_recovery` test 

https://github.com/paritytech/polkadot-sdk/actions/runs/15958913569/job/45008816489

causing it to fail

### Steps to reproduce the issue

Issue occurs in a random manner. Hard to reproduce.

### Describe the results you received

In this case the reason is that eve and validator-2 try to use the same port:
```
❯ rg 45739 pr_err1/
pr_err1/logs/eve.log
48:2025-06-29 20:12:41.128  INFO tokio-runtime-worker sub-libp2p: [living-bread-1409] 🔍 Discovered new external address for our node: /ip6/::1/tcp/45739/ws/p2p/relaynode_eve
52:2025-06-29 20:12:41.161  INFO tokio-runtime-worker sub-libp2p: [living-bread-1409] 🔍 Discovered new external address for our node: /ip4/127.0.0.1/tcp/45739/ws/p2p/relaynode_eve
53:2025-06-29 20:12:41.434  INFO tokio-runtime-worker sub-libp2p: [living-bread-1409] 🔍 Discovered new external address for our node: /ip4/172.18.0.2/tcp/45739/ws/p2p/relaynode_eve

pr_err1/logs/validator-2.log
29:2025-06-29 20:12:42.592 ERROR tokio-runtime-worker prometheus: Error binding to '0.0.0.0:45739': Os { code: 98, kind: AddrInUse, message: "Address already in use" }
```

`eve` spawned with command:
```
[...]

PR #368, opened 2025-06-30
Fix port assignment

lrubasze:
This PR is supposed to fix https://github.com/paritytech/zombienet-sdk/issues/367

It makes sure all ports being used are parked during NodeSpec generation.

cc: @pepoviola


pepoviola:
Thanks!!

Issue #383, opened 2025-07-17
Sharing node names between relay chain and parachain doesn't work


bkchr:
### Issue Description

```
Caused by:
    Duplicated node name: alice)
```

Before it was possible to have some network definition where a node could be for example called `alice` for the relay chain and the parachains.

### Steps to reproduce the issue

Steps to reproduce the issue
1.
2.
3.

### Describe the results you received

Describe the results you received

### Describe the results you expected

Describe the results you expected

### Zombienet version

master

### Provider

Native

### Provider version

```yaml
## For binaries
polkadot 0.9.40-a2b62fb872b
polkadot-parachain 0.9.380-fe24f39507f 

## For Kubernetes/Podman
podman version 4.4.1

OR

kubectl version v0.26.3
cluster version 1.25.2
```

### Upstream Latest Release

Yes

### Additional environment details

Additional environment details

### Additional information

Additional information like issue happens only occasionally or issue happens with a particular architecture or on a particular setting

PR #391, opened 2025-08-08
Implement logic to deduplicate node names.

RafalMirowski1:
https://github.com/paritytech/zombienet-sdk/issues/383


RafalMirowski1:
Noticed now, I think I should probably also add deduplication when adding to the running network.

Issue #217, opened 2024-05-16
[config] change relay/para builders to be more explicit and less confusing.


pepoviola:
Currently we have `with_node` (in relay) and `with_collator` (in para) but the `node` is by default a _validator_ and can be confusing. To reduce the friction we should:

In relay builder:
**add**:
`with_validator` (make _authority_, pass --validator flag)
**change**:
 `add_node` (_full node_)

In para builder:
**add**:
 `add_node` (_full node_)
**change**:
`with_collator` (make _authority_, pass --collator flag)

And remove the methods that allow to set the _role_ of the node.

---

cc: @bkchr

PR #419, opened 2025-09-19
Change relay/para builders to be more explicit and less confusing

DenzelPenzel:
#217


crates/configuration/src/parachain.rs
```diff
@@ -833,6 +833,8 @@ impl<C: Context> ParachainConfigBuilder<WithId, C> {
             self.default_chain_context(),
             self.validation_context.clone(),
         ))
+        .validator(true)
+        .with_args(vec!["--collator".into()])
```
pepoviola, line 837: If the `node` is set as validator we inject automatically the `--collator` flag.
```suggestion
```

crates/configuration/src/parachain.rs
```diff
@@ -892,6 +896,41 @@ impl<C: Context> ParachainConfigBuilder<WithAtLeastOneCollator, C> {
         }
     }
 
+    /// Add a new full node using a nested [`NodeConfigBuilder`].
+    /// The node will be configured as a full node (non-validator).
+    pub fn add_node(
```
pepoviola, line 901: maybe we can rename this to `with_fullnode` so the name is very clear about the type of node.
wdyt?

crates/examples/examples/common/lib.rs
```diff
@@ -12,8 +12,8 @@ pub fn small_network_config(
             r.with_chain("rococo-local")
                 .with_default_command("polkadot")
                 .with_default_image("docker.io/parity/polkadot:latest")
-                .with_node(|node| node.with_name("alice"))
-                .with_node(|node| node.with_name("bob"))
+                .with_validator(|node| node.with_name("alice"))
+                .with_fullnode(|node| node.with_name("bob"))
```
pepoviola, line 16: why we change this (and the others) to be fullnodes? I think we should keep as validators and if you want include a new _fullnode_ extra

pepoviola (commented):
Looks good!! a few comments and is ready to merge 🚀

pepoviola (approved):

PR #424, opened 2025-09-29
Fix: Race in unit tests - Verify default spawn timeout before setting…

mchristou:
If ```default_node_spawn_timeout_works_when_env_is_set``` is executed before ```default_node_spawn_timeout_falls_back_to_default_when_env_is_not_set``` it fails because the environment variable is already set.


pepoviola (commented):
small nit

PR #434, opened 2025-10-20
refactor: remove pjs feature and related code

### LINKED ISSUE #433: [pjs] Remove pjs as dep
Let's remove for now `pjs-rs` as dep since we need a reliable way to update _deno_ crates without causing issues. We can then incorporate as extension or something similar.

DenzelPenzel:
close #433


pepoviola (approved):
Looks good, thanks!!

PR #441, opened 2025-10-27
feat: produce the chainspec in a local file

### LINKED ISSUE #440: Improve chain-spec builder
After talk with @moliholy about #438, we should allow to create a chain-spec with _cmds_ that not use `stdout` and just write the spec to a file. For that I think the best approach is to allow to _optionally_ set the expected file output in the `CommandInContext`

```
pub enum CommandInContext {
    Local(String),
    Remote(String),
}
```

So, we can run the`cmd` and pick the destination file to continue with the current process.

cc: @moliholy

moliholy:
Closes #440.

This PR allows specifying a `chain_spec_command_output_path` parameter in the toml configuration file that indicates where the `chain_spec_command` has produced the output. `/dev/stdout` remains the default.

It also opens up the door to generating any artifact so that the command writes into a file, like the genesis code or the initial state, but for now no extra parameters have been added for those, yet it'd be straightforward to add them.

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -248,7 +248,12 @@ impl ChainSpec {
             //     check the [`DEFAULT_PRESETS_TO_CHECK`] in order to find one valid
             // - If we can't find any valid preset use the `default config` from the runtime
 
-            let preset = DEFAULT_PRESETS_TO_CHECK
+            let preset_to_check = if let Some(preset) = &runtime.preset {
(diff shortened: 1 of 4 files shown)


pepoviola:
> > Maybe is good also to generate a test that use this feature.
> 
> I created two [tests](https://github.com/paritytech/zombienet-sdk/pull/441/files#diff-36b22e43fb705e442dde7678d7d46d7285c23b86246a143dd36481795542056eR236) that check a dummy JSON was produced in a file and also test the original /dev/stdout execution path keeps working. Do you have more test cases in mind?

No, is ok with this tests.
Thx!

pepoviola (approved):

PR #449, opened 2025-10-29
fix: actually run command locally

moliholy:
This PR is more of a discussion rather than something I'm hoping to get merged. Apparently, I misunderstood the purpose of "local" chain-spec command. It seems that it's simply not using a foreign provider (docker, pod, kubernettes, etc).

My initial assumption, and the reason I was so excited for, is that the flag would allow to execute the command _in the current directory the user is in_, presumably the parachain project folder. By executing it in a temporary directory we loose the context, and forces parachain developers to have in-house binaries that already contain the runtime (like old-fashioned nodes), which does not play nice with omni-node-based parachains.

The purpose of the changes proposed in this PR is simply to really execute "locally" the command. Not just in the local host, but in the directory the user is in currently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10361,7 +10361,7 @@ dependencies = [
 
 [[package]]
 name = "zombie-cli"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "clap",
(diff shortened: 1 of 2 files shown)


moliholy:
@pepoviola awesome! Many many thanks!

Issue #444, opened 2025-10-29
Publish crates on new releases


pepoviola:
We need to add the ci machinery to auto-publish the crates to crates.io when we create a new release.
Thx!

PR #452, opened 2025-10-31
CI: Publish crates on new releases

mchristou:
Implementation of #444 

For now the dry-run flag is added until we have a successful dry-run


pepoviola (approved):
looks good, thanks!

PR #453, opened 2025-11-04
feat: add session key support

DenzelPenzel:
close https://github.com/paritytech/zombienet/issues/1906


crates/configuration/src/shared/node.rs
```diff
@@ -158,6 +160,12 @@ impl Serialize for NodeConfig {
         state.serialize_field("p2p_port", &self.p2p_port)?;
         state.serialize_field("p2p_cert_hash", &self.p2p_cert_hash)?;
 
+        if let Some(session_key) = &self.session_key {
```
pepoviola: We can just use this attr
```
#[serde(skip_serializing_if = "Option::is_none")]
```

Issue #456, opened 2025-11-11
Impl logs and dump_logs fns in docker provider


pepoviola:
Currently are unimplemented with `todo!()`s

Thx!

PR #457, opened 2025-11-11
Impl logs and dump_logs fns in docker provider

RafalMirowski1:
closes https://github.com/paritytech/zombienet-sdk/issues/456

### DIFF
--- a/crates/file-server/src/main.rs
+++ b/crates/file-server/src/main.rs
@@ -65,7 +65,7 @@ async fn upload(
         }
 
         let stream = request.into_body().into_data_stream();
-        let body_with_io_error = stream.map_err(io::Error::other);
+        let body_with_io_error = stream.map_err(|err| io::Error::new(io::ErrorKind::Other, err));
         let body_reader = StreamReader::new(body_with_io_error);
         futures::pin_mut!(body_reader);
 


crates/provider/src/docker/client.rs
```diff
@@ -529,6 +529,41 @@ impl DockerClient {
...
+
+        // most useful logs are in stderr so we need to merge both
+        let mut lines: Vec<_> = String::from_utf8_lossy(&output.stdout)
+            .lines()
+            .chain(String::from_utf8_lossy(&output.stderr).lines())
+            .map(|s| s.to_string())
+            .collect();
+
+        // sort by prefix (should be timestamp)
+        lines.sort_by_key(|line| line.split_whitespace().next().unwrap_or("").to_string());
```
pepoviola: I think we can somehow redirect stderr to stdout (`2>&1`) or something similar to avoid re-order.

pepoviola: I just changed this to use pipes and don't need to re-order after read.
thx!

crates/provider/src/docker/client.rs
```diff
@@ -529,6 +529,39 @@ impl DockerClient {
         Ok(containers)
     }
 
+    pub(crate) async fn container_logs(&self, container_name: &str) -> Result<String> {
```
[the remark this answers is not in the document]
pepoviola: hm, yes. That could happen with big logs (and also re-order could be costly). I think we can explore 2 options:

1. Use a wrapper with`sh -c "docker log..."` that allow us to redirect in the args __but__ we rely on the system `sh` (that most of the time is present but could cause some edge case).
2. Use a file redirection, and the just read the file and return. Is a little bit extra work but could work.

I think (2) is better but (1) will work most of the times too (99% I will say)

wdyt?

Cargo.toml
```diff
@@ -17,7 +17,7 @@ members = [
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
 version = "0.4.2"
-rust-version = "1.70.0"
+rust-version = "1.87.0"
```
pepoviola: I think this isn't need anymore.

```suggestion
rust-version = "1.70.0"
```

PR #463, opened 2025-11-17
feat: chain specification generation

DenzelPenzel:


crates/examples/examples/chain_spec_runtime_kusama.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let temp_dir = Builder::new()
+        .prefix("zombienet-chain-spec-")
+        .tempdir()
+        .context("creating temporary directory")?;
+
+    let temp_path = temp_dir.path().to_path_buf();
+
+    let kusama_runtime_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../orchestrator/testing/kusama_runtime-v2000002.compact.compressed.wasm");
```
pepoviola: I think we can get these from the release page directly without have them in our repo.

crates/orchestrator/src/generators/chain_spec.rs
```diff
@@ -392,147 +392,224 @@ impl ChainSpec {
         };
 
         if (self.runtime.is_some() && self.asset_location.is_none()) || is_omni_node {
```
pepoviola: We should remove this workaround. The idea is to follow this flow:
- First try to create the raw spec using `GenericChainSpec`
- IFF fails with the expected errors 
  - Fallback to use the cmd logic

PR #470, opened 2025-11-28
feat: implement run_script for Docker provider

### LINKED ISSUE #469: [docker] Implement `run_script` for docker provider
https://github.com/paritytech/zombienet-sdk/blob/main/crates/provider/src/docker/node.rs#L506

We should allow to run script as we do with k8s, uploading the script (make it executable), run it and get the output.

Thx!

mchristou:
Closes #469


crates/orchestrator/src/network/node.rs
```diff
@@ -273,6 +273,22 @@ impl NetworkNode {
...
 
+    /// Run a script inside the node's container/environment
+    ///
+    /// The script will be uploaded to the node, made executable, and executed with
+    /// the provided arguments and environment variables.
+    ///
+    /// Returns `Ok(stdout)` on success, or `Err((exit_status, stderr))` on failure.
+    pub async fn run_script(
+        &self,
+        options: provider::types::RunScriptOptions,
```
pepoviola: Let's bring those to the scope with
```rs
use provider::types::{RunScriptOptions, ExecutionResult};
```

```suggestion
        options: RunScriptOptions,
```

crates/provider/src/docker/node.rs
```diff
@@ -501,9 +501,30 @@ where
...
-        todo!()
+        let file_name = options
+            .local_script_path
+            .file_name()
+            .expect(&format!(
+                "file name should be present at this point {THIS_IS_A_BUG}"
+            ))
+            .to_string_lossy();
+
+        let remote_script_path = PathBuf::from(format!("/tmp/{file_name}"));
```
pepoviola: Is better to use the `scripts_dir` from the node (_provider node_) itself. (https://github.com/paritytech/zombienet-sdk/blob/main/crates/provider/src/docker/node.rs#L104)

PR #484, opened 2025-12-19
Fix genesis-state generator args handling

mchristou:
**Problem:** 
Args passed via `with_genesis_state_generator` were ignored and the subcommand was duplicated.

**Fix:**
Always append generator args (skip duplicate subcommand) and make selection precedence explicit: path → generator → default.

Added unit tests for precedence and for preserving generator args.
Added example.

issue spotted here: https://github.com/paritytech/polkadot-sdk/actions/runs/20345159002/job/58456898092?pr=10651


mchristou:
Hi @pepoviola 
Do you think we need similar changes for `with_genesis_wasm_generator` ?

pepoviola (approved):

mchristou:
> Looks good @mchristou, can you check the clippy error? Thx!

Thanks @pepoviola
I fixed the error in different PR.

PR #497, opened 2026-01-15
(feat): Design decorators

### LINKED ISSUE #496: Design _decorators_ for sdk
We want to support some _kind_ of decorators, as we support in [v1](https://github.com/paritytech/zombienet). They are mostly used to allow users to customize the logic we use to customize the chain-spec. In the previous version we _host_ the decorators so users need to contribute to use to use them or use zombienet from a _modified_ source code. This cause some friction in the past, so for this version we want to develop some kind of _plugins_ machinery that allow users to set where to load them and make them available without need to change the main code.
One possible path, is to allow to set the _decorators path_ in the config, and we use that directory to load/call the scripts/wasm with some _well-known_ naming.

In general we want to allow this fns to be customized:

- clear_authorities
- add_authorities
- add_aura_authorities
- add_collator_selection
- add_balances
- add_staking
- add_grandpa_authorities
- add_hrmp_channels

And also add some fns (one for paras one for relay) to allow a final customization.

Thx!

mchristou:
Closes #496 

Introduce decorator system for dynamic chain-spec customization

Adds```DecoratorRegistry``` and ```ChainSpecDecorator``` trait that enable runtime-specific chain-spec modifications without modifying source code.

Optional hooks to override default behavior:
  
  - clear_authorities
  - add_authorities
  - add_aura_authorities
  - add_collator_selection
  - add_balances
  - add_staking
  - add_grandpa_authorities
  - add_hrmp_channels

High level customization: ```customize_para``` and ```customize_relay```
Example: ```decorators_usage.rs```

 **Benefits:**
  
  - Dynamic modifications at runtime
  -  No source code changes required
  -  Backwards compatible (falls back to defaults when no decorator provided)

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -937,7 +937,7 @@ impl ChainSpec {
         hrmp_channels: &[HrmpChannelConfig],
         para_artifacts: Vec<ParaGenesisConfig<U>>,
         scoped_fs: &ScopedFilesystem<'a, T>,
-        decorator_registry: Option<&crate::decorators::DecoratorRegistry>,
+        decorator_registry: Option<&DecoratorRegistry>,
     ) -> Result<(), GeneratorError>
     where
         T: FileSystem,


pepoviola:
Hi @mchristou, thanks for check this one. The issue that this approach doesn't play well with the _cli_, since we want to allow users to define a directory where we will load the _plugins_. Maybe we can just rely in this first version on call scripts, passing the chain-spec content as argument and read the output. wdyt?

PR #500, opened 2026-01-16
Implement script option to run on top of chain-spec generation

mchristou:
Implements #496 

Add per-chain post-process script support

Adds optional post_process_script field to parachain/relaychain configs. Script runs on the plain chain-spec before raw generation with a single path argument.

### DIFF
--- a/crates/examples/examples/configs/chain_spec_command.toml
+++ b/crates/examples/examples/configs/chain_spec_command.toml
@@ -1,0 +1,21 @@
+[relaychain]
+chain = "rococo-local"
+default_command = "polkadot"
+post_process_script = "crates/examples/scripts/customize-chain-spec.sh"
+
+[[relaychain.nodes]]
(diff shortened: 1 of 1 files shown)


crates/examples/examples/chain_spec_command.rs
```diff
@@ -0,0 +1,144 @@
...
+        .with_env_filter(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        )
+        .init();
+
+    let script_path = concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/scripts/customize-chain-spec.sh"
```
pepoviola: I think this file is missing.

pepoviola (commented):
Looks good! can we add an example using the _cli_ (with toml config).
Thx!

Issue #501, opened 2026-01-19
Dynamically added collators incorrectly appear in relay chain node list


DenzelPenzel:
### Issue Description

When dynamically adding a collator to a running parachain network using Network::add_collator(), the collator node is incorrectly added to both the relay chain's node list and the parachain's collator list.

### Steps to reproduce the issue

Steps to reproduce the issue
1. Deploy a network with a relay chain and parachain (example `small_network_with_para.rs`)
2. Dynamically add a new collator using `network.add_collator("new-col-1", options, 100).await?`
3. Inspect the generated `zombie.json` file 
4. Observe that the collator appears in both:
       - .relay.nodes[] array (incorrect)
       - .parachains."100"[0].collators[] array (correct)

### Describe the results you received

 `add_collator()` method incorrectly passes `None` as the `para_id` parameter to `add_running_node()`:
`self.add_running_node(node, None).await;  // should be Some(para_id)`

### Describe the results you expected

Collator should only appear in the parachain's collator list:

Relay chain nodes: alice, bob, new1 (3 nodes)
Parachain 100 collators: collator, new-col-1 (2 nodes)

### Zombienet version

v0.4.4

### Provider

Native

### Provider version

```yaml
polkadot 1.20.0-15a3abd6ccb
polkadot-parachain 1.20.0-15a3abd6ccb
```

### Upstream Latest Release

Yes

### Additional environment details

Additional environment details

PR #503, opened 2026-01-19
Prevent dynamically added collators from being added to relay chain node list

DenzelPenzel:
#501 

 Fixes incorrect network topology in `zombie.json`


pepoviola:
Thanks!!

PR #515, opened 2026-02-09
Observability add-on (Grafana + Prometheus) for native provider

DenzelPenzel:
- Add opt-in observability stack that spawns Prometheus + Grafana as Docker/Podman containers, auto-configured to scrape all nodes metrics endpoints
  - Enabled via [settings.observability] in TOML config the stack starts automatically during network spawn

### DIFF
--- a/crates/sdk/tests/two-paras-same-id.rs
+++ b/crates/sdk/tests/two-paras-same-id.rs
@@ -1,7 +1,7 @@
 use zombienet_sdk::{environment::get_spawn_fn, NetworkConfigBuilder};
 
 #[tokio::test(flavor = "multi_thread")]
-async fn two_paras_same_id() {
+async fn ci_two_paras_same_id() {
     tracing_subscriber::fmt::init();
     let spawn_fn = get_spawn_fn();
     let config = NetworkConfigBuilder::new()


crates/orchestrator/src/network.rs
```diff
@@ -826,6 +838,67 @@ impl<T: FileSystem> Network<T> {
...
+    /// network.start_observability(&obs_config).await?;
+    /// println!("Grafana: {}", network.observability().unwrap().grafana_url);
+    /// #   Ok(())
+    /// # }
+    /// ```
+    pub async fn start_observability(
+        &mut self,
+        config: &configuration::ObservabilityConfig,
+    ) -> Result<&ObservabilityInfo, anyhow::Error> {
+        if self.observability.is_none() {
```
pepoviola: Maybe we can track the status of the observability in the network.

crates/orchestrator/src/observability.rs
```diff
@@ -0,0 +1,242 @@
...
+        if let Some(port) = extract_port_from_uri(&node.prometheus_uri) {
+            targets.push_str(&format!(
+                "      - targets: ['{host_addr}:{port}']\n        labels:\n          node: '{}'\n",
+                node.name
+            ));
+        }
+    }
+
+    format!(
+        "global:\n  scrape_interval: 15s\n  evaluation_interval: 15s\n\nscrape_configs:\n  - job_name: 'zombienet'\n    metrics_path: /metrics\n    static_configs:\n{targets}"
```
pepoviola: maybe scrape every 5s

pepoviola (commented):
looks good @DenzelPenzel, a few comments inline.

crates/orchestrator/src/network.rs
```diff
@@ -826,6 +838,67 @@ impl<T: FileSystem> Network<T> {
...
+            config,
+            &nodes,
+            self.ns.name(),
+            self.ns.base_dir(),
+            &self.filesystem,
+        )
+        .await?;
+
+        self.observability = Some(info);
+        Ok(self.observability().unwrap())
```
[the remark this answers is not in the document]
DenzelPenzel: Removed unwrap

crates/orchestrator/src/network.rs
DenzelPenzel: Tracking observability status in network

DenzelPenzel:
> Looks good, we should also update the _docs_ to refer to this new feature. Can you do it as follow up pr?

Sure!

PR #528, opened 2026-03-24
Add `restart_with()` to `NetworkNode`

tdimitrov:
I'm writing a zombienet-sdk test which tests a version downgrade. In nutshell it starts a network with binary version N and some command line arguments A1, validates it works then restarts each node with binary version N-1 and command line arguments A2. E.g.:

```rust
async fn downgrade_validators_test() -> Result<(), anyhow::Error> {
	let _ = env_logger::try_init_from_env(
		env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "info"),
	);

	let config = build_network_config()?;
	let network = initialize_network(config).await?;

	let relay_client: zombienet_sdk::subxt::OnlineClient<PolkadotConfig> =
		network.get_node("validator-0")?.wait_client().await?;

	// Assert network is working before downgrade.
	log::info!("Asserting para throughput before downgrade");
	assert_para_throughput(&relay_client, 20, [(ParaId::from(PARA_ID), 10..25)]).await?;
	log::info!("Network is operating correctly with {}", POLKADOT_IMAGE_RC);

	// Restart all validators with the downgraded image.
	log::info!("Restarting all validators with image {}", POLKADOT_IMAGE_DOWNGRADE);
	for name in VALIDATOR_NAMES {
		let node = network.get_node(*name)?;
		node.restart_with(
			Some("/Users/ceco/parity/src/psdk-v1.21.2/target/testnet/polkadot".to_string()),
			Some(vec![("-lparachain=debug").into()]),
			None,
			None,
		)
		.await?;
		log::info!("Restarted {} with {}", name, POLKADOT_IMAGE_DOWNGRADE);
	}

	// Reconnect to validator-0 after restart.
	let relay_client: zombienet_sdk::subxt::OnlineClient<PolkadotConfig> =
		network.get_node("validator-0")?.wait_client().await?;

	// Assert network is still working after downgrade.
	log::info!("Asserting para throughput after downgrade");
	assert_para_throughput(&relay_client, 20, [(ParaId::from(PARA_ID), 10..25)]).await?;
	log::info!("Network is operating correctly with {}", POLKADOT_IMAGE_DOWNGRADE);

	log::info!("Test finished successfully");
	Ok(())
}
```

To achieve this I needed a `restart_with` function which restarts the node by overriding it's command and/or arguments. Since this function is missing I've added an implementation for it.

# What's changed
Add the `restart_with` function to the `ProviderNode` trait:

```rust
async fn restart_with(
        &self,
        program: Option<String>,
        args: Option<Vec<String>>,
        image: Option<String>,
        after: Option<Duration>,
    ) -> Result<(), ProviderError>;
```
All the parameters are `Option<..>`. If they are `Some(..)` the corresponding parameter is overridden. If `None` - the initial value is kept. The function is implemented for all providers:
- native -> I've tested it manually for the native provider.
- docker -> not yet tested, but I will test it by myself.
- k8s -> not tested and not sure how to test it by myself.

# Additional changes
For the implementation I needed to make some additional changes.

## internal_args and user_args
Besides the parameters passed by the user with the network config, zombienet-sdk adds additional parameters indicating db paths, ports to be used, etc. Since these parameters must not be touched I've separated the arguments in two fields:
- `user_args`, pased by the user, can be changed with `restart_with`
- `internal_args`, set by the SDK, not to be changed by the user

## `args` now returns a copy of the command line arguments
Since `user_args` is now behind a `RwLock` returning a reference to the arguments is not possible. For this reason `args()` (from `ProviderNode`) now returns `Vec<String>` which is a copy of all arguments used by the binary.

# Note about k8s provider implementation
Since my k8s knowledge is very limited this part is mostly vibe coded. I went through all the changes and from my point of view they are sane but I'd appreciate extra eyes. 

The most significant change here is the introduction of persistent volumes, so that the DB can survive the POD recreation. This change was suggested by claude, sounded sane to me so I kept it. Please check if it does make sense.

# Note about testing
As mentioned, I've tested the native provider manually and will test the docker one too. For k8s - I am open for suggestions.

Also any ideas about automated tests I could have added are welcome.


pepoviola:
Hey @tdimitrov, thanks for this feature. I think we can simplify the approach by constrain the way they supply the new binary version (so, we don't need any extra coordination for docker/k8s). I will try to sketch a simplify version of this and  ping you.

tdimitrov:
Implemented in https://github.com/paritytech/zombienet-sdk/pull/531

PR #542, opened 2026-05-15
[sketch] API for snapshot generation

michalkucharczyk:
This is just proposal to agree the shape of the API.

### DIFF
--- a/sketch.rs
+++ b/sketch.rs
@@ -19,16 +19,17 @@ impl NetworkNode {
 }
 
 /// Identity-handling policy for [`NetworkNode::snapshot_db`].
+#[derive(Default)]
 pub enum SnapshotKind {
-    /// Includes `keystore/` and `network/`. Safe to load back into a
-    /// single node. If consumed by multiple sibling nodes the shared
-    /// session keys cause equivocation and the libp2p identity causes
-    /// peer-id collisions.
-    Full,
-    /// Strips `keystore/` and `network/`. Safe to load on any number of
-    /// sibling nodes; zombienet re-injects per-node keys at startup via
-    /// `author_insertKey`.
+    /// Strips the node's identity, only DB. Safe to load on any number of sibling
+    /// nodes.
+    #[default]
     Shareable,
+    /// Includes the node's identity (`keystore/` and `network/`). Use
+    /// only when the snapshot will be loaded back into a node that plays
+    /// the same role as the source. Loading a `Full` snapshot on a
+    /// different node causes equivocation and peer-id collisions.
+    Full,
 }
 
 /// Result of [`NetworkNode::snapshot_db`].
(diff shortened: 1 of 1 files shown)


sketch.rs
```diff
@@ -0,0 +1,114 @@
...
+    ) -> Result<NodeSnapshot, anyhow::Error>;
+}
+
+/// Identity-handling policy for [`NetworkNode::snapshot_db`].
+pub enum SnapshotKind {
+    /// Includes `keystore/` and `network/`. Safe to load back into a
+    /// single node. If consumed by multiple sibling nodes the shared
+    /// session keys cause equivocation and the libp2p identity causes
+    /// peer-id collisions.
+    Full,
```
skunert: Do you have a specific use-case in mind for this? I think for existing tests I have not encountered this yet.

michalkucharczyk: No. But I can imagine that such scenario (replicating 1:1 entire source network) will be needed. API can be ready for this.

pepoviola: In general we always inject the keystore (derived from the node name), but we can also add support to not doit iff we have this kind of snap. Maybe set `Shareable` as default?
Thx!

sketch.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub fn add(self, snap: NodeSnapshot) -> Self;
+
+    /// Attach an arbitrary serializable blob. Stored as JSON under
+    /// `user_data` in the manifest. Test authors put block heights,
+    /// CIDs, release tags, "number of collators", etc. here.
+    pub fn user_data<T: serde::Serialize>(self, data: T) -> Self;
+
+    /// Produce `out_path` (gzipped tarball). Fails if no archives were
+    /// added.
+    pub fn build(self, out_path: impl AsRef<Path>) -> Result<Bundle, anyhow::Error>;
```
pepoviola: Maybe we can use the same _type state_ pattern and only allow to call build if _add_ was called before. wdyt?

sketch.rs
michalkucharczyk: One more comment expaining why I proposed Shared / Full:

The problem with snapshot that was fixed by https://github.com/paritytech/smoldot/pull/3264 was as follows:
- snapshot contain the _keystore_ and _networking_  directory,
- the same snapshot was used for alice and bob nodes,
- this led to problems with the liveliness of the network  (basically network was not progressing).

This approach was incorrect.

So when generating the snapshot we can take two approaches:
- we always snapshot DB directory only, and always re-generated the keys (consensus + p2p),
OR:
- we also allow to take a "full" snapshot for every node individually and then replicate the nodes from the snapshot (so we have per-node snapshot). This could provide some means for scenarios when we want to have full control over the node's directory content.

For now we only have use-cases for _shared_ snapshots.

sketch.rs
michalkucharczyk: 0a04d26

sketch.rs
michalkucharczyk: > Maybe set Shareable as default?

0a04d26

pepoviola: I think the issue (in smoldot), was a bug on the _native_ provider (https://github.com/paritytech/zombienet-sdk/blob/main/crates/provider/src/native/node.rs#L189-L195), since we are _unpacking_ the snap as last step before initialize the process. In the other providers (docker/k8s), we first unpack the snap and then apply the _startup files (e.g keystore).

I will fix this and ping you.

(https://github.com/paritytech/zombienet-sdk/issues/543)

michalkucharczyk: But still - the keys should not be there in the snapshot. We should only have DB there if we don't plan to use keys/networking.

We can also skip the snapshot type in the API. What would be your take on this?

PR #544, opened 2026-05-26
snapshot cache added: race avoided

michalkucharczyk:
There is a race condition in initialize_db_snapshot when a db_snapshot `AssetLocation` is shared across multiple sibling nodes. Each spawned node independently downloads/copies the asset into the namespace cache at `{ns_base_dir}/{sha256(loc)}.tgz`. This happens to be executed in parallel.

 The current workaround in downstream tests (e.g. smoldot [`smoke_generate_snapshots.rs`](https://github.com/paritytech/smoldot/blame/7fbf4e0e898d235a2c25cf4b0f5cd74cbac6e45b/e2e-tests/tests/smoke_generate_snapshots.rs#L75-L83)) is to pre-stage per-node copies under unique filenames so the cache keys don't collide.

This PR fixes it by moving asset resolution upstream of the parallel fanout:
  - New `orchestrator::generators::resolve_db_snapshots` walks every `NodeSpec::db_snapshot`, deduplicates by `AssetLocation`, and serially populates the cache. Runs once in `Orchestrator::spawn_inner` (and once per `Network::add_node / add_collator / add_parachain`) before any node  spawns.
  - `SpawnNodeOptions::db_snapshot` flips from `Option<AssetLocation>` to `Option<PathBuf>` - the resolved cache path. Each provider's `initialize_db_snapshot` now takes a `&Path` and only extracts; no downloads, no exists() check, no race.
  - `ZOMBIE_RM_TGZ_AFTER_EXTRACT` cleanup moves from the per-node code path (would delete the cache while sibling nodes are still reading it) to `cleanup_db_snapshot_cache`, called after all consumers finish.

 Public SDK API (`with_db_snapshot`, `AssetLocation`, etc.) is unchanged.

### DIFF
--- a/.github/workflows/ci_integration.yml
+++ b/.github/workflows/ci_integration.yml
@@ -81,7 +81,7 @@ jobs:
         run: |
           export ZOMBIE_K8S_CI_NAMESPACE=$(cat /data/namespace)
           mkdir -p /tmp/zombie-1/logs
-          
+
           # Install kubectl if not available
           if ! command -v kubectl &> /dev/null; then
             echo "Installing kubectl..."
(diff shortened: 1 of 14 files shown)


michalkucharczyk:
This is some clean-up work required for implementation of https://github.com/paritytech/zombienet-sdk/pull/542

pepoviola (approved):
Looks good, thanks @michalkucharczyk !!

pepoviola:
Hi @michalkucharczyk, thanks for the fix 🙌

Issue #541, opened 2026-05-15
Add support for generating snapshot for given network


michalkucharczyk:
### Is your feature request related to a problem? Please describe.

Would be nice to have support for snapshot generation directly in zombienet-sdk.
E.g. `node.get_snapshot()` which can later be used for starting from snapshot.

I will provide more details about API soon.

### Describe the solution you'd like

I will provide more details about API soon.

michalkucharczyk:
@pepoviola some proposal here: 
https://github.com/paritytech/zombienet-sdk/pull/542/

Let's discuss there if shape is OK.

PR #545, opened 2026-05-27
snapshot api added

michalkucharczyk:
Implementation of proposal from #541 (also #542).

Tests that build DB snapshot artifacts (cumulus `full_node_warp_sync`, smoldot `smoke`/`bulletin`) each hand-roll the same pause → tar → checksum → bundle dance. This moves it into the SDK.

### What's new

- **`NetworkNode::snapshot_db(out_path)`** — tars a node's DB into a `.tgz` (`data/`, plus `relay-data/` for cumulus collators). Drops `keystore/` and `network/` so the archive can be loaded on several sibling nodes without key/peer-id clashes.
- **`Network::pause()` / `resume()`** — SIGSTOP/SIGCONT all nodes in parallel, so the DB on disk is consistent while you snapshot.
- **`snapshot::BundleBuilder`** — packs per-node archives + a JSON `user_data` blob into one `bundle.tar.gz` with a `manifest.json` (sha256 + size per archive). `build()` won't compile until you've added at least one archive.
- **`with_optional_default_db_snapshot(Option<…>)`** on the relay/parachain builders (and `with_optional_db_snapshot` per node) — so one network builder works for both the fresh and the from-snapshot case without branching.

### Example

```rust
// Produce: snapshot a relay validator + a collator, bundle them.
network.pause().await?;
let relay = network.get_node("alice")?.snapshot_db("relay.tgz").await?;
let para  = network.get_node("collator")?.snapshot_db("para.tgz").await?;
network.resume().await?;

let bundle = BundleBuilder::new()
    .add(relay)
    .add(para)
    .user_data(serde_json::json!({ "para_best_block": 42 }))
    .build("bundle.tar.gz")?;          // bundle.path / .sha256 / .size

////////////////////////////////////////////////////////////////////////////////

// Consume: same builder, snapshot optional.
fn network(snapshot: Option<&Path>) -> NetworkConfig {
    NetworkConfigBuilder::new()
        .with_relaychain(|r| {
            r.with_chain("rococo-local")
                .with_optional_default_db_snapshot(snapshot)   // None = fresh, Some = resume
                .with_validator(|n| n.with_name("alice"))
        })
        /* … */
        .build().unwrap()
}
```

### Tests

- Unit tests in `crates/sdk/src/snapshot.rs`: bundle build → unpack → manifest round-trip (no network, runs in CI).
- `crates/configuration`: tests for the three `with_optional_*db_snapshot` methods.
- `crates/sdk/tests/snapshot_roundtrip.rs` (`#[ignore]`, needs polkadot binaries): snapshot a live network, bundle it, re-spawn from the snapshot, and check both chains advance and finalize past the snapshot height.

## Notes

- `snapshot_db` reads the node's local base dir, so it's native-provider only.
- Caller must pause the node first — snapshoting a running node risks a torn RocksDB state.
- A graceful `terminate()` (instead of `pause`) before snapshotting is left for a follow-up.
- Open question if `SIGSTOP` allows for safe db copy. So far it worked well.


crates/orchestrator/src/network/node.rs
```diff
@@ -333,6 +335,84 @@ impl NetworkNode {
...
+    /// the produced archive is safe to load on any number of sibling nodes.
+    ///
+    /// Operates on the node's local base dir, so it's meaningful for the
+    /// native provider only. The caller is responsible for pausing the node
+    /// first; snapshotting a running node risks a torn RocksDB state.
+    pub async fn snapshot_db(
+        &self,
+        out_path: impl AsRef<Path>,
+    ) -> Result<NodeSnapshot, anyhow::Error> {
+        let out_path = out_path.as_ref().to_path_buf();
```
pepoviola, line 353: This logic should be in the `native` provider and use `inner` from here to call it. We can let k8s/docker as _unimplemented_ for now (but is also possible to implement).

Thx!

crates/sdk/src/snapshot.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// `user_data` in the manifest. Test authors put block heights,
+    /// CIDs, release tags, "number of collators", etc. here. Can be
+    /// called before or after [`add`](BundleBuilder::add); last call wins.
+    pub fn user_data<T: Serialize>(mut self, data: T) -> Self {
+        self.user_data = serde_json::to_value(&data).unwrap_or(serde_json::Value::Null);
+        self
+    }
+}
+
+fn build_bundle(
```
pepoviola, line 161: Just _nit_, since we provide a nice way to generate the bundle maybe we can also include the `untar_bundle` fn here to not allow users to use it directly and not need to re-implement.
wdyt?

pepoviola:
Awesome job @michalkucharczyk !! I added a couple of comments, if you are agree I can make those changes and also wire your test to run in CI.

Thx!

michalkucharczyk:
> Awesome job @michalkucharczyk !! I added a couple of comments, if you are agree I can make those changes and also wire your test to run in CI.
> 
> Thx!

Yes, feel free to take it from here. But I can also implement your requests. Whatever works for you :)

crates/orchestrator/src/network/node.rs
```diff
@@ -333,6 +335,84 @@ impl NetworkNode {
...
+            .await
+            .map_err(|err| anyhow!("spawning tar for node {}: {err}", self.name))?;
+        if !status.success() {
+            return Err(anyhow!(
+                "tar failed for node {} (status {status})",
+                self.name
+            ));
+        }
+
+        let bytes = tokio::fs::read(&out_path)
```
skunert: We should not read t he entire file here and hash it, but do this iteratively to save some memory.

skunert (commented):
Nice!

pepoviola:
ping @michalkucharczyk, I applied the changes and wire the test to run in ci. If you are ok, I will merge it and draft a new release.
Thx!

crates/provider/src/native/node.rs
```diff
@@ -729,6 +735,70 @@ where
...
+        })?;
+
+        if !status.success() {
+            return Err(ProviderError::SnapshotDb(
+                self.name().into(),
+                anyhow!("tar failed for node {} (status {status})", self.name),
+            ));
+        }
+
+        let mut file = File::open(&out_path).unwrap();
```
michalkucharczyk:
```suggestion
        let mut file = File::open(&out_path).map_err(|err| ProviderError::SnapshotDb(self.name().into(), anyhow!(err)))?;
```

crates/sdk/src/snapshot.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    let gz = tar.into_inner().context("finishing tar")?;
+    let mut f = gz.finish().context("finishing gzip")?;
+    f.flush().context("flushing bundle file")?;
+    drop(f);
+
+    let bytes = std::fs::read(&out_path)
+        .with_context(|| format!("reading produced bundle {}", out_path.display()))?;
+    let size = bytes.len() as u64;
+    let sha256 = hex::encode(sha2::Sha256::digest(&bytes));
```
michalkucharczyk: stream could be used here too (not reading entire file to memory).

crates/sdk/tests/snapshot_roundtrip.rs
```diff
@@ -0,0 +1,324 @@
...
+    timeout: u64,
+) -> anyhow::Result<()> {
+    node.wait_metric_with_timeout(BEST_BLOCK, move |x| x >= best, timeout)
+        .await?;
+    node.wait_metric_with_timeout(FINALIZED_BLOCK, move |x| x >= finalized, timeout)
+        .await?;
+    Ok(())
+}
+
+fn untar_bundle(bundle_path: &Path, out_dir: &Path) -> anyhow::Result<()> {
```
michalkucharczyk: nit: maybe this could be used in tests?

crates/provider/src/native/node.rs
pepoviola: ahh, let me remove the `unwrap` 🤦‍♂️

michalkucharczyk:
Nit: maybe we could keep `NetworkNode::base_dir()` ? (e.g. for inspection during test authoring?)

crates/sdk/src/snapshot.rs
pepoviola: yes, I will apply the same logic that in the _node_

crates/sdk/tests/snapshot_roundtrip.rs
pepoviola: You mean moving to the snapshot module?

crates/orchestrator/src/network/node.rs
```diff
@@ -333,6 +334,44 @@ impl NetworkNode {
...
+            }
+        );
+
+        let InnerSnapshotDb {
+            filename,
+            sha256,
+            size,
+        } = self.inner.snapshot_db(is_cumulus_based).await?;
+
+        // now we need to _move_ the inner file to the out_path
```
michalkucharczyk, line 368: nit: It acutally copies the file. Maybe this could be avoided by tzg'ing the file to the destination location? Or moving instead copying?

pepoviola: Yes, could be an optimization for _native_ but the idea here is that this method _can_ be used in all the supported providers, so in docker/k8s we need to get the file from the pod. 

Thx!

PR #551, opened 2026-06-20
feat(mcp): add read-only zombie-mcp diagnostics server

DenzelPenzel:
Adds `zombie-mcp`, a read-only diagnostics tool for a Zombienet run you already
started. It uses `zombie.json` as the run handle and inspects configs, logs,
liveness, metrics, and block production — it never spawns or destroys networks.

It works two ways, sharing one core:

- **CLI (deterministic, no LLM)** — run the diagnostics by hand or in CI.
- **MCP server (additive)** — let an AI agent (Codex / Claude Code) drive the
  same core via natural language, correlate evidence with repo source, and
  reason about errors outside the baked checks.

## Core tools / checks

- `find_recent_runs` — discover recent run candidates with no arguments
- `diagnose_run` — triage a started run from `zombie_json_path`, including
  scanning logs near the expected path when startup failed before `zombie.json`
  was written
- `validate_config` — validate a network config file
- `list_nodes` / `get_node_logs` — inspect an attached live network
- `check_node_liveness` / `check_block_production` — node health checks
- `query_metric` — read a Prometheus metric from a node

## CLI mode

```sh
zombie-mcp diagnose --auto                       # find newest run and diagnose
zombie-mcp diagnose --zombie-json <path>         # diagnose a specific run
zombie-mcp validate  --config <toml>             # validate a config
```

The JSON `DiagnosticReport` goes to stdout; logs to stderr. The MCP server is an
optional `mcp` feature (on by default) — core + CLI build without it
(`--no-default-features`).

## Install (MCP)

```sh
zombie-mcp install codex --force      # Codex
zombie-mcp install claude --force     # Claude Code
```

Then, after a run fails: *"Debug my zombienet run with zombie-mcp."*

## Diagnostics correctness

- `check_block_production` no longer counts the genesis block (#0) as
  production — a stalled chain times out instead of a false success.
- A node that is up but stuck at best block 0 is flagged (`no_block_progress`)
  instead of being reported healthy.
- `next_steps` covers the stalled-block case (registration / connectivity /
  para id) rather than "No immediate failure detected".
- Identical findings (e.g. a log error repeated across relaunch attempts) are
  collapsed so the summary is honest.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11064,7 +11064,7 @@ dependencies = [
 
 [[package]]
 name = "zombie-mcp"
-version = "0.4.13"
+version = "0.4.14"
 dependencies = [
  "anyhow",
  "clap",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5068,6 +5068,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "nix"
+version = "0.30.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74523f3a35e05aba87a1d978330aef40f67b0304ac79c1c00b294c9830543db6"
+dependencies = [
+ "bitflags 2.10.0",
+ "cfg-if",
+ "cfg_aliases",
+ "libc",
+]
+
+[[package]]
 name = "nodrop"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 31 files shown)


pepoviola:
Hey @DenzelPenzel, thanks for your contribution. What's the reasoning behind this tool? the idea is just allow cloude/codex to allow to attach and run those _diagnosis_ functions? can you include some example of the usage in the _pr_  also, there are several parts in the code that _needs to improve_ :)

Thx!

pepoviola (changes requested):
Thanks for your contribution, let's touch base to discus some design in this tool.

PR #552, opened 2026-06-20
Correct the Core Count Check

0xOmarA:
# Description

This is a small PR which corrects the core count & validator count checks we have. We used to require that we have more validators than cores which I think is the wrong check to have, we need to check that we have as many validators as cores (or more).

This has been tested with retester to make sure that networks can be spawned (and operate correctly) when validators = number of cores and networks operate correctly.

### DIFF
--- a/crates/orchestrator/src/generators/session_0_overrides.rs
+++ b/crates/orchestrator/src/generators/session_0_overrides.rs
@@ -111,7 +111,7 @@ pub fn generate_session_0_overrides(
 
     // some checks first
     if num_genesis_cores > session.validators.len() as u32 {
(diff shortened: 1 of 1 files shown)


pepoviola (approved):
Thanks!!

0xOmarA:
@pepoviola Any chance we can cut a new release with this small fix so I can depend on this new version in retester rather than depending on a git dependency? :)

pepoviola:
> @pepoviola Any chance we can cut a new release with this small fix so I can depend on this new version in retester rather than depending on a git dependency? :)

Hey @0xOmarA, yes :) 

New version (`v0.4.14` ) is already in crates.io

Thx!!

PR #557, opened 2026-07-12
Use export-chain-spec as the default chain-spec command

Kanasjnr:
## Summary
- Change `DEFAULT_CHAIN_SPEC_TPL_COMMAND` from deprecated `build-spec` to `export-chain-spec`
- Drop `--disable-default-bootnode` from the default template (`export-chain-spec` does not support it; zombienet still customizes bootnodes after generation)

Needed so polkadot-sdk can remove `build-spec` without every test overriding `chain_spec_command` (see paritytech/polkadot-sdk#12586).

## Test plan
- [ ] Existing zombienet-sdk CI passes
- [ ] Spawning a network without a custom `chain_spec_command` runs `<binary> export-chain-spec --chain <name>`

### DIFF
--- a/crates/orchestrator/src/shared/constants.rs
+++ b/crates/orchestrator/src/shared/constants.rs
@@ -8,9 +8,11 @@ pub const RPC_PORT: u16 = 9944;
 pub const RPC_HTTP_PORT: u16 = 9933;
 // P2P default port
 pub const P2P_PORT: u16 = 30333;
-// default command template to build chain-spec
+// Default command template to export a chain-spec.
+// `export-chain-spec` replaced the deprecated `build-spec` CLI in polkadot-sdk.
+// Bootnodes are customized after generation, so `--disable-default-bootnode` is unused.
 pub const DEFAULT_CHAIN_SPEC_TPL_COMMAND: &str =
-    "{{mainCommand}} build-spec --chain {{chainName}} {{disableBootnodes}}";
+    "{{mainCommand}} export-chain-spec --chain {{chainName}}";
 // interval to determine how often to run node liveness checks
 pub const NODE_MONITORING_INTERVAL_SECONDS: u64 = 15;
 // how long to wait before a node is considered unresponsive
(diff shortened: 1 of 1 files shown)


crates/orchestrator/src/shared/constants.rs
```diff
@@ -8,9 +8,11 @@ pub const RPC_PORT: u16 = 9944;
 pub const RPC_HTTP_PORT: u16 = 9933;
 // P2P default port
 pub const P2P_PORT: u16 = 30333;
-// default command template to build chain-spec
+// Default command template to export a chain-spec.
+// `export-chain-spec` replaced the deprecated `build-spec` CLI in polkadot-sdk.
+// Bootnodes are customized after generation, so `--disable-default-bootnode` is unused.
```
bkontur: I think this added docs does not say much relevant.
```suggestion
// Default command template to export a chain-spec.
```

pepoviola:
Hey @Kanasjnr, sorry about the delay and thanks for the contribution. Did you know from which version this subcommand (export-chain-spec) is available? Since we need to keep support for people using old versions, they can still customize the command but maybe we can just detect old versions and switch back to the `build-spec` one.

Thx!

Kanasjnr:
> Hey @Kanasjnr, sorry about the delay and thanks for the contribution. Did you know from which version this subcommand (export-chain-spec) is available? Since we need to keep support for people using old versions, they can still customize the command but maybe we can just detect old versions and switch back to the `build-spec` one.
> 
> Thx!

Hey @pepoviola no worries, thanks for taking a look.

`export-chain-spec` was added in paritytech/polkadot-sdk#7719 . From what I can tell it’s present from `polkadot-stable2506` onwards; `stable2503` and earlier still only have `build-spec`.

So yeah, detecting older binaries and falling back to `build-spec` sounds good. i can adjust the PR that way if you want e.g. try `export-chain-spec` by default, and if the binary doesn’t support it (or version < stable2506), use `build-spec` instead. People can still override `chain_spec_command` either way.

Kanasjnr:
Closed it has been resolved via https://github.com/paritytech/zombienet-sdk/pull/561

PR #561, opened 2026-08-11
add logic to use export-chain-spec or build-spec as subcommand

pepoviola:
replace #557, add logic to use `export-chain-spec` when is present and fallback to `build-spec` for old nodes.

cc: @bkontur / @Kanasjnr

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -154,15 +154,27 @@ impl ChainSpec {
                 .first()
                 .expect("main_cmd should be preset to build the expect. qed");
             debug!("resolving subcommand for: {main_cmd}");
-            let help_output = ns
-                .get_node_available_args((main_cmd.to_string(), self.image.clone()))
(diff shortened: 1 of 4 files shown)


bkontur:
> add logic to use `export-chain-spec` when is present and fallback to `build-spec` for old nodes.

@pepoviola how old nodes do we want to support? This export command was introduced in https://github.com/paritytech/polkadot-sdk/pull/7719 and it is part of stable2506., isn't this enough? Is anybody using older nodes than that? Otherwise, we don't get rid of that `build-spec` command

Kanasjnr:
Hmm.. @bkontur  i think the fall back approch is actually good because zombienet-sdk has a broader user base than just polkadot-sdk, teams running older testnets or non-Polkadot chains may still have pre-stable2506 binaries. 
Once this merges, I can remove the per-test with_chain_spec_command overrides in [#12586](https://github.com/paritytech/polkadot-sdk/pull/12586) and close #557 as superseded. 

LGTM on the approach, @pepoviola.

PR #562, opened 2026-08-14
feat: add functions to load config from TOML string

GCrispino:
- Add functions to load config from TOML string. All of them take TOML-formatted string data and return `Result<NetworkConfig, anyhow::Error>`. These are the functions:
  - `load_from_toml_string_with_base_dir(toml_text: &str, base_dir: &Path)` - parses the TOML string and takes a base directory, such that relative paths in the config are replaced using this directory as base. This was implemented here because that's the logic already present in the `load_from_toml` function, which uses the path of the directory of the file path it receives
  - `load_from_toml_string(toml_data: &str)` - same as `load_from_toml_string_with_base_dir`, but using `std::env::current_dir()?` as base directory
- Refactor `load_from_toml` - it now uses `load_from_toml_string_with_base_dir` after loading the file contents from the received file path;
- Bump `sc-chain-spec` dependency from `v50.0.0` to `v51.0.0`(fixes dependency issues related to yanked `core2` crate)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
- "crypto-common",
+ "crypto-common 0.1.6",
  "generic-array",
 ]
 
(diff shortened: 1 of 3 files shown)


pepoviola (commented):
Looks good 👍

pepoviola (approved):

pepoviola:
@GCrispino Is this ready to review/merge?

PR #571, opened 2026-08-26
fix session0 overrides, when there are not cores to assign

pepoviola:
cc: @GCrispino / @sandreim 

Add check to ensure not panic when users set to override session0 __but__ there isn't paras registered in genesis.

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -2136,6 +2136,12 @@ mod tests {
     }
 
     #[test]
+    fn merge_with_empty_b_works() {
+        let mut chain_spec_json = chain_spec_with_stake();
+        merge(&mut chain_spec_json, &json!({}));
+    }
+
+    #[test]
     fn get_min_stake_works() {
         let mut chain_spec_json = chain_spec_with_stake();
 
--- a/crates/orchestrator/src/generators/session_0_overrides.rs
+++ b/crates/orchestrator/src/generators/session_0_overrides.rs
@@ -3,6 +3,7 @@ use codec::{Decode, Encode};
 use serde_json::json;
 use sp_core::crypto::AccountId32;
 use support::substorage::storage_value_key;
+use tracing::warn;
 
 use crate::generators::errors::GeneratorError;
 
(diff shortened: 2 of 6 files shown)


crates/orchestrator/src/generators/session_0_overrides.rs
```diff
@@ -77,6 +78,10 @@ pub fn generate_session_0_overrides(
     num_genesis_cores: u32,
 ) -> Result<serde_json::Value, GeneratorError> {
     let mut overrides = json!({});
+    if num_genesis_cores == 0 {
+        warn!("'num_genesis_cores' is 0, means that we can not override session_0. Please check your config to ensure you have paras to register in chain-spec.");
+        return Ok(overrides);
+    }
```
GCrispino, line 84: This looks good to avoid that we don't actually try to generate overrides without any parachain configuration.

However, maybe something to look in the future, I suppose that a lot of other code in lib.rs (where `generate_session_0_overrides` is called) can also be skipped in the case that we have no parachains

crates/orchestrator/src/generators/session_0_overrides.rs
```diff
@@ -77,6 +78,10 @@ pub fn generate_session_0_overrides(
     num_genesis_cores: u32,
 ) -> Result<serde_json::Value, GeneratorError> {
     let mut overrides = json!({});
+    if num_genesis_cores == 0 {
+        warn!("'num_genesis_cores' is 0, means that we can not override session_0. Please check your config to ensure you have paras to register in chain-spec.");
```
GCrispino: minor thing: shouldn't `session_0` be `session 0`?

crates/orchestrator/src/generators/chain_spec.rs
```diff
@@ -2135,6 +2135,12 @@ mod tests {
         }})
     }
 
+    #[test]
+    fn merge_with_empty_b_works() {
+        let mut chain_spec_json = chain_spec_with_stake();
+        merge(&mut chain_spec_json, &json!({}));
+    }
+
```
GCrispino, line 2143: is this related to this PR?

GCrispino (approved):
leaving some minor comments but approving in case you'd like to merge as is

crates/orchestrator/src/generators/chain_spec.rs
pepoviola: Since we may call `merge` with an empty object as _b_ side, a test to ensure that works cover that.

crates/orchestrator/src/generators/session_0_overrides.rs
pepoviola: Yes, we can add a main check there to not follow that path. I will add it now.

crates/orchestrator/src/generators/chain_spec.rs
GCrispino: How does that relate to the changes in the current PR?

pepoviola: calling merge with empty _b side_ is one of the paths this pr includes.

crates/orchestrator/src/generators/session_0_overrides.rs
GCrispino: Thanks!

PR #572, opened 2026-08-26
specify https in wget calls in helper-binaries-downloader script

GCrispino:
`wget` calls in the [helper-binaries-downloader.sh](https://github.com/paritytech/zombienet-sdk/blob/main/crates/provider/src/shared/scripts/helper-binaries-downloader.sh) script to download `curl` and `coreutils` were timing out (in docker mode, at least).

If we specify the `https://` prefix in these calls, then the downloads work. This PR adds that to the script


pepoviola (approved):
thanks!

Issue #566, opened 2026-08-18
[jam] Chain-spec generation


pepoviola:
Integrate the logic to generate and customize the chain-spec for `jam`

Issue #565, opened 2026-08-18
[jam] Configuration -> Spec


pepoviola:
Wire the configuration to the generation of the spec that allow to spawn a _jam_ chain.

pepoviola:
impl in #573

PR #573, opened 2026-08-27
Jam basic integration

### LINKED ISSUE #564: [jam] Configuration
We want to add support to represent a jam chain as part of the configuration:
- Exploring if we can rely in the current set of fields or needs to add a new _discriminant_ one.
- If we need a new filed, make the design to easily extend it.
- Create the transition to read form `toml` and emit a config.

pepoviola:
Initial iteration of jam integration:
- Allow to setup a jam chain with the builder
- Allow to use a predefined tiny setup directly
- Support load a `toml` config with the _zombie-cli_
- __ONLY__ works with _native_ provider at the moment.

cc: @michalkucharczyk / @sandreim 

fix #564 #565 #566


michalkucharczyk:
Would it be possible to build JAM network with given service binary?

I have fake [_parasim_](https://github.com/paritytech/parachain-service/blob/mku-jam-collator-poc/testing/parasim-service/src/lib.rs) which mimics the _parachain-service_ and emulates the parachain progress on JAM.

It would be perfect if I could specify something like:
```rust
jam_builder.with_service(path_to_jam_service_binary)
```

The service could be added into JAM genesis -- I never tried it but maybe it is doable? 

I am currently using the following command to deploy it:
```
jamt create-service ./parasim-service.jam 1000000000000000 --register=parasim --raw --id 7
```

In future we could just switch to real parachain-service and start testing different scenarios.

WDYT? If reasonable, would you be willing to check this path?

pepoviola:
https://github.com/paritytech/zombienet-sdk/pull/573#issuecomment-5452707232

ping @zdave-parity, since this will be used for _testnets_ I think allow adding a service in genesis could speed things and also remove the `jamt` dependency (at least for spawn). Maybe we can add only for `dev` (or custom chain ids), wdyt?

zdave-parity:
> [#573 (comment)](https://github.com/paritytech/zombienet-sdk/pull/573#issuecomment-5452707232)
> 
> ping @zdave-parity, since this will be used for _testnets_ I think allow adding a service in genesis could speed things and also remove the `jamt` dependency (at least for spawn). Maybe we can add only for `dev` (or custom chain ids), wdyt?

`dev` has a fixed chainspec. This would only make sense for custom chains with custom chainspecs. The main reason I had for not wanting to support this is that it would force us to improve the tooling around creating services etc. If that tooling is bypassed then there is no motivation to make it good :). But if it really would be significantly more convenient to just have services included in genesis state then we can add support to PolkaJam's `gen-spec` for this. Jan was proposing a change to how service code is built which would probably require something like this anyway (see https://github.com/paritytech/polkajam/pull/1211). I'm assuming here that you are using `gen-spec` to generate your chainspecs, or are you doing something else?

On a related note, I guess it would be useful to expose the `gen-spec` stuff in library form? This code could probably be pulled out of PolkaJam into its own crate. Or possibly just added to `jam-std-common`. I think it only really needs `jam-std-common` stuff, not any internal node stuff.

michalkucharczyk:
If `gen-spec` allows to build custom spec (e.g. containing the service) which can later be used for start the jam local testnet, then yes - this is what we look for.

crates/orchestrator/src/generators/jam_config.rs
```diff
@@ -0,0 +1,87 @@
...
+                Some(GenesisValidator {
+                    peer_id: n.peer_id.clone(),
+                    bandersnatch: n
+                        .accounts
+                        .accounts
+                        .get("bandersnatch")
+                        .expect("bandersnatch should be present.")
+                        .public_key
+                        .clone(),
+                    net_addr: format!("127.0.0.1:{}", n.rpc_port.0),
```
michalkucharczyk: should be p2p port?

michalkucharczyk:
```suggestion
                    net_addr: format!("127.0.0.1:{}", n.port.0),
```

crates/orchestrator/src/spawner.rs
```diff
@@ -390,3 +391,183 @@ pub async fn spawn_process(
...
+    // )?;
+
+    // let prometheus_uri = format!("http://{ip_to_use}:{prometheus_port_external}/metrics");
+    info!("🚀 {}, should be running now", node.name);
+    info!(
+        "💻 {}, peer details {}@{ip_to_use}:{}",
+        node.name, node.peer_id, node.port.0
+    );
+    if node.mode == JamNodeMode::Ordinary {
+        info!("💻 {}, rpc  {ip_to_use}:{}", node.name, node.port.0);
```
michalkucharczyk:
```suggestion
        info!("💻 {}, rpc  {ip_to_use}:{}", node.name, node.rpc_port.0);
```

pepoviola:
Thanks for your comments @zdave-parity / @michalkucharczyk, at the moment zombienet is _calling_ `gen-spec` command to build a custom chain-spec based on a custom config (with the validators info). We can customize the generated spec to include a service as we do with polkadot __but__ if we can include the `gen-spec` as library I think will be a more clean/better approach.

bkchr:
> `dev` has a fixed chainspec. This would only make sense for custom chains with custom chainspecs. The main reason I had for not wanting to support this is that it would force us to improve the tooling around creating services etc. If that tooling is bypassed then there is no motivation to make it good :). But if it really would be significantly more convenient to just have services included in genesis state then we can add support to PolkaJam's `gen-spec` for this.

We should register this at genesis. Otherwise we will waste CI cycles waiting for the service to be registered and whatever. Also this would not really improve the tooling, because it would be hidden somewhere deep inside of zombienet.

michalkucharczyk:
We could have a `GenSpecBuilder` public API, something like this.
This could cut entire overhead of using other tools to deploy and configure the service.

```rust
    GenSpecBuilder::new(id, validators)
      .bootstrap_service(blob)              // what gen-spec hardcodes now, made explicit
      .service(id, blob, ServiceParams { balance, min_item_gas, min_memo_gas, deposit_offset })
      .service_storage(id, key, value)      // bumps items/bytes footprint
      .preimage(host_service_id, blob)      // = solicit+provide at slot 0 (e.g. authorizer)
      .auth_queue(core, authorizer_hash)    // pools derive from queues at genesis
      .assigner(core, service_id)           // Privileges.assign[core]
      .privileges(...)                      // bless/designate/register/always_acc, for GENESIS.md
      .bootnodes(...)
      .build() -> ChainSpec
```

zdave-parity:
> Also this would not really improve the tooling, because it would be hidden somewhere deep inside of zombienet.

It would presumably use `jamt` or a library version of that.

> We could have a `GenSpecBuilder` public API, something like this. This could cut entire overhead of using other tools to deploy and configure the service.
> 
> ```rust
>     GenSpecBuilder::new(id, validators)
>       .bootstrap_service(blob)              // what gen-spec hardcodes now, made explicit
>       .service(id, blob, ServiceParams { balance, min_item_gas, min_memo_gas, deposit_offset })
>       .service_storage(id, key, value)      // bumps items/bytes footprint
>       .preimage(host_service_id, blob)      // = solicit+provide at slot 0 (e.g. authorizer)
>       .auth_queue(core, authorizer_hash)    // pools derive from queues at genesis
>       .assigner(core, service_id)           // Privileges.assign[core]
>       .privileges(...)                      // bless/designate/register/always_acc, for GENESIS.md
>       .bootnodes(...)
>       .build() -> ChainSpec
> ```

Happy to accept a PR for something like this in the polkajam repo. Re this specific proposal:

- I would call it `ChainSpecBuilder`. `GenSpecBuilder` is kind of redundant.
- It should not have a `bootstrap_service` method. The bootstrap service is not special, it should be added the same way as other services.

As I mentioned above, this can probably just go in `jam-std-common`. We should rework the `gen-spec` command etc to use this rather than duplicating the logic.

michalkucharczyk:
> Happy to accept a PR for something like this in the polkajam repo

I am on it. Will push something soon.

pepoviola:
Merging now, the integration with the chain-spec lib will land in followup prs.

Issue #578, opened 2026-08-28
Fix build-spec for test-parachain bin


pepoviola:
see err https://github.com/paritytech/polkadot-sdk/pull/13019#issuecomment-5452580657

The binary needs to use `build-spec` as subcmd.

PR #583, opened 2026-09-02
Stop hardcoding build-spec for test-parachain

Kanasjnr:
## Summary
- Remove the `test-parachain` / empty-chain `polkadot-parachain` special case that always forced `build-spec`.
- Always resolve the subcommand from binary `--help`, preferring `export-chain-spec` when present and falling back to `build-spec` otherwise (same path as #561).

## Why
#579 / #578 hardcoded `build-spec` for `test-parachain`. That worked while those binaries still exposed `build-spec`, but [polkadot-sdk#12586](https://github.com/paritytech/polkadot-sdk/pull/12586) removes `build-spec`, so zombienet CI fails with:

```
error: unexpected argument 'build-spec' found
Usage: test-parachain [OPTIONS] [-- <RELAYCHAIN_ARGS>...]
```

Without this fix, polkadot-sdk [cannot drop `build-spec` ](https://github.com/paritytech/polkadot-sdk/pull/12586)without restoring per-test `with_chain_spec_command` overrides for every `test-parachain` network.

### DIFF
--- a/crates/orchestrator/src/generators/chain_spec.rs
+++ b/crates/orchestrator/src/generators/chain_spec.rs
@@ -349,16 +349,30 @@ impl ChainSpec {
                 }
             };
 
+            self.resolve_subcommand(ns).await?;
+
             // SAFETY: we ensure that command is some with the first check of the fn
             // default as empty
             let sanitized_cmd = if replacement_value.is_empty() {
-                // we need to remove the `--chain` flag
-                self.command.as_ref().unwrap().cmd().replace("--chain", "")
+                if self.subcommand.as_deref() == Some("export-chain-spec") {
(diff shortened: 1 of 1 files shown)


bkchr (approved):

pepoviola (approved):

pepoviola (approved):

Kanasjnr:
Thanks for merging this! 

One follow-up: this fix is currently only on main (workspace version still shows 0.4.18, same as the last published release, cut in #582 before this fix). polkadot-sdk's #12586 (removing the deprecated build-spec CLI subcommand) is blocked on a zombienet-sdk release that includes this fix, since polkadot-sdk pins an exact crates.io version.

Could you help cut a new version that includes this commit? @pepoviola

PR #584, opened 2026-09-03
Implement toml support for JAM

bkchr:


crates/configuration/src/network.rs
```diff
@@ -171,6 +180,17 @@ impl NetworkConfig {
...
         // All unwraps below are safe, because we ensure that the relaychain is not None at this point
         if network_config.relaychain.is_none() {
+            // A JAM chain replaces the relaychain, none of the validation below applies.
+            if network_config.jamchain.is_some() {
+                if !network_config.parachains.is_empty() {
+                    // TODO: parachains on top of a JAM chain aren't wired up yet.
+                    Err(anyhow!(
+                        "Parachains are not supported yet with a JAM chain."
+                    ))?
+                }
```
pepoviola: We are working on this with @michalkucharczyk, and currently works ok for _parasim_ parachain. Maybe we can drop this check for now.

crates/orchestrator/src/generators/command.rs
```diff
@@ -481,6 +472,22 @@ pub fn generate_for_jam_node(node: &JamNodeSpec, options: GenCmdOptions) -> (Str
         cmd_args.push(tel_endpoint.into())
     }
 
+    // Args set in the config, appended last so they can't clash with the ones above.
```
pepoviola, line 484:
```suggestion
    // Args set in the config, appended last so they can't clash with the ones above.
    // TODO: ensure that we are not overwriting the args zombienet add automatically.
```

crates/orchestrator/src/lib.rs
```diff
@@ -214,6 +213,22 @@ where
...
+        if network_spec.try_relaychain().is_none() {
+            let jam_spec =
+                network_spec
+                    .jamchain
+                    .as_ref()
+                    .ok_or(OrchestratorError::InvalidConfig(
+                        "A network needs either a relaychain or a jamchain.".to_string(),
+                    ))?;
+
+            return self
```
pepoviola, line 227: We want to still spawn paras, I can wire this 👍

pepoviola:
Thanks @bkchr! If you are ok I can remove the `spawn paras constraint` before merge it.

crates/configuration/src/network.rs
```diff
@@ -2284,6 +2276,97 @@ command = "polkadot"
...
 
+    #[test]
+    fn jamchain_toml_without_relaychain_works() {
+        let config = NetworkConfig::load_from_toml_string(
+            r#"
+[jamchain]
+id = "dev"
+default_command = "polkajam"
+
+[[jamchain.nodes]]
```
pepoviola, line 2287: Make sense to validate the chain params match? the default is `tiny` that needs 6 validators, so we can invalidate a configuration with a diff count of validators?

pepoviola (approved):
thanks!!!

